PR #272, opened 2020-09-09
ACMS-000: Remove unnecessary dependency on CKEditor

wimleers:
…. Drop the ckeditor dependency from acquia_cms_common.

See https://github.com/acquia/cohesion-dev/commit/b9cfe12 🤓


vishalkhode1:
Closing seems it's no longer required, as it's in open state since a very long time.

PR #690, opened 2021-03-05
ACMS-587: Configuration of backstop using a js settings file

saurabhtripathics:
**Motivation**
We should be able to run the backstop and pa11y commands from IDE.
Fixes #690 

**Proposed changes**
- Added code to separate out the commands for IDE.

**Alternatives considered**
-  NONE

**Testing steps**
- Head towards terminal on your PC and run the following commands to test.
- npm run backstop-starter - Runs on both IDE and local
- npm run pa11y-starter
- npm run pa11y-starter-ide

**Merge requirements**
- [* ] Manual testing by a reviewer

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,10 +34,12 @@
     "install-tools": "npm ci",
     "uninstall-tools": "rm -r node_modules;",
     "pa11y-starter": "pa11y-ci --config tests/.pa11yci.js",
-    "backstop-starter": "backstop test --config=tests/backstop/backstop.json"
+    "pa11y-starter-ide": "pa11y-ci --config tests/.pa11yci.js http://127.0.0.1:8000/",
+    "backstop-starter": "backstop test --config=tests/backstop/backstop-settings.js"
   },
   "dependencies": {
     "backstopjs": "^5.0.6",
+    "minimist": "^1.2.5",
     "pa11y-ci": "^2.4.0",
     "scripts": "^0.1.0"
   }
--- a/tests/backstop/backstop-settings.js
+++ b/tests/backstop/backstop-settings.js
@@ -1,3 +1,41 @@
+/*
+  How to use
+
+  backstop reference --configPath=backstop-settings.js
+       backstop test --configPath=backstop-settings.js
+
+  backstop reference --configPath=backstop-settings.js --refhost=http://example.com
+       backstop test --configPath=backstop-settings.js --testhost=http://example.com
+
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -34,10 +34,13 @@
     "install-tools": "npm ci",
     "uninstall-tools": "rm -r node_modules;",
     "pa11y-starter": "pa11y-ci --config tests/.pa11yci.js",
+    "pa11y-starter-ide": "pa11y-ci --config tests/.pa11yci.js http://127.0.0.1:8000/",
+    "backstop-starter-ide": "backstop test --config=tests/backstop/backstop-settings.js --testhost=http://127.0.0.1:8000/ --pathfile=paths --labelfile=labels",
```
vishalkhode1, line 38: @panshulK Is there any need to pass `testhost` argument from here as we've already added condition check in `backstop-settings.js` to default use `http://127.0.0.1:8000` baseUrl.

tests/backstop/backstop-settings.js
```diff
@@ -0,0 +1,121 @@
...
+    "delay": 0,
+    "hideSelectors": [],
+    "removeSelectors": [],
+    "hoverSelector": "",
+    "clickSelector": "",
+    "postInteractionWait": 0,
+    "selectors": [ 'document' ],
+    "selectorExpansion": true,
+    "expect": 0,
+    "misMatchThreshold": 10.0,
```
vishalkhode1, line 117: @panshulK In backstop json config `tests/backstop/backstop.json` which runs in local has different values for different scenarios. But ideally I think it should use same vales everywhere i.e IDE, local, travis etc.

tests/backstop/backstop-settings.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+// The host to test
+if (!arguments.testhost) {
+  arguments.testhost  = "http://127.0.0.1:8000"; // Default test host
+}
+// The host to reference
+if (!arguments.refhost) {
+  arguments.refhost  = "http://127.0.0.1:8000"; // Default test host
+}
+/*
```
vishalkhode1, line 36: @panshulK Since we've already added default refhost & testhost url, so this should work in IDE, local as well as in travis. So, instead of managing two files i.e `backstop.json` & this file i.e `backstop-settings.js` . Can't we remove backstop.json file and incorporate changes from that file here ? 

I know ticket mentions to use separate command  but I think we can avoid this.

cc: @chandan-singh7929

PR #746, opened 2021-03-25
ACMS-707: Filter type are not getting filtered properly when filtered via Facets

chandan-singh7929:
**Motivation**

_Facet Pretty Path Default working:_
Facet pretty path by default adds tid to the term while creating its alias. For e.g. `Meet Up` term alias will be `meet-123` where 123 is the term id. While decoding the alias into term, it takes the tid and use it for filtering purpose. 

_ACMS Customisation_
In ACMS, for the clean URLs, the default beahviour is further modified to remove the `tid` from the alias and do a database query to obtain the term id. So, according to Acquia CMS `Meet Up` term alias will be `meet` and then a database query is used to obtain the term which should be an exact match of meet.

_Issue_
Since ACMS removes the tid from the alias, and database query is an exact match query. `Meet` never matches to `Meet Up` and hence no results are obtained.

Fixes #NNN

**Proposed changes**

- Remove custom facet pretty path plugin
- Use the default coder provided by facet pretty path module.

**Alternatives considered**

- Do not use pathauto's cleanstring, rather create own cleanString method to clean the terms for pretty paths. 
- Instead of exact database search, to a pattern match. Process the terms return by pattern by using cleanString method and then return the term id.

**Testing steps**

- Go to events view and click on ,meet-up filter, and check if results are being fetched
- Verify all view pages and check each term provided in facet pretty path block and make sure all are working as exptect.

**Merge requirements**
- [_Bug_] _Major change_, _Minor change_, _Bug_, _Enhancement_, and/or _Chore_ label applied
- [x] Manual testing by a reviewer


modules/acquia_cms_common/tests/src/ExistingSite/BreadcrumbTest.php
```diff
@@ -162,15 +162,15 @@ private function assertBreadcrumb(array $expected_breadcrumb) : void {
       if ($link) {
         return [
           $link->getText(),
-          $link->getAttribute('href'),
+          // Let remove -tid from the link generated by facet pretty path.
+          preg_replace('/-[^-]*$/', '', $link->getAttribute('href')),
```
vishalkhode1, line 166: @chandan-singh7929 Isin't this would return wrong url, if URL contains multiple hyphens. 
Ex: /meet-up/something-123. This would return => /meet
cc: @gauravgoyal

chandan-singh7929: @vishalkhode1 - I have tested this, it will only remove string from last part of the URL in this case(`something-123`), not from the beginning or middle

chandan-singh7929:
All test case has been fixed with additional coverage for video embed. please have a look.
cc: @gauravgoyal @vishalkhode1 
Ran test suite locally here is the result:
Screenshot 2021-03-30 at 8 17 05 PM

gauravgoyal (approved):

PR #775, opened 2021-04-20
ACMS-709: Change pipeline config removing gz files from being committ…

saurabhtripathics:
…ed to the pipelines artifact
https://backlog.acquia.com/browse/ACMS-709
**Motivation**
Removing gz files from being committed to the pipelines artifact should clear up the space, which is what we need.

**Proposed changes**
Change pipeline config

**Alternatives considered**
pruning the Acquia Cloud repo

**Testing steps**
 Need to find out, i think when pipeline jobs will run, will verify cloud repo, it should not have these gz files.


vishalkhode1:
Closing this for now. If we need it, will create a separate ticket.

PR #779, opened 2021-04-21
ACMS-734: Change dir for siteinstallcommand to avoid it getting calle…

saurabhtripathics:
…d twice

**Motivation**
Rebuild is getting called twice 
it is because drupal finds two post-hook command i.e

- ./drush/Commands/SiteInstallCommands.php
- ./docroot/profiles/acquia_cms/drush/Commands/SiteInstallCommands.php

**Proposed changes**
 
 Move folder drush/Commands to src. 
So, file SiteInstallCommands.php will be moved from drush/Commands/SiteInstallCommands.php to src/Commands/SiteInstallCommands.php

 

**Testing steps**
 in acms and acms-project rebuild should get called only once after site install


saurabhtripathics:
closing this one, will create new branch

PR #791, opened 2021-05-04
ACMS-620: setup subtree splitter using shell script and run split com…

saurabhtripathics:
…mands
https://backlog.acquia.com/browse/ACMS-620
**Motivation**
The goal is for Acquia CMS development to still happen in the main acquia_cms github repo. The script in (2) below should be run periodically to push commits from the main acquia_cms repo's module folders to each module's individual repos. For example, if development work is done on the acquia_cms_common module in the main repo, those commits should be pushed into the individual acquia_cms_common repo via this script. Running the script manually should suffice.
**Proposed changes**
1. Use Git Subtree splitter tool (https://github.com/splitsh/lite) to split the main acquia_cms repo’s modules folder content into individual repos. Installing the binary should be easiest.

2. Create a bash script to push commits from the main acquia_cms repo into the split repos. See: https://github.com/laravel/framework/blob/7.x/bin/split.sh
**Alternatives considered**

**Testing steps**
On dev environment we will need to test,
that if we do a commit in source repo, than it gets traversed/updated to destination repo


acms_split.sh
```diff
@@ -0,0 +1,79 @@
...
+echo -e "${GREEN} Script Running on ${OSTYPE}${NOCOLOR}"
+
+# Sample usage: ./runsplit.sh
+# Commit will be traversed/updated from source to destination repositary.
+
+# Dynamically setting current branch to checked out branch.
+# For example if this script is run on branch, test than,
+# commit wil traverse from test branch of source repo to,
+# test branch of destination repo.
+CURRENT_BRANCH=`git rev-parse --abbrev-ref HEAD`
```
saurabhtripathics, line 21: I think we will need to change this after dev testing? Either we can pass this as parmeter to script, or provide a static value like develop. 
Currently we are dynamically setting current branch to checked out branch.
For example if this script is run on branch, test than,
commit will traverse from test branch of source repo to,
test branch of destination repo.

acms_split.sh
```diff
@@ -0,0 +1,77 @@
...
+echo -e "${GREEN} Script Running on ${OSTYPE}${NOCOLOR}"
+
+# Sample usage: ./runsplit.sh
+# Commit will be traversed/updated from source to destination repositary.
+
+# Dynamically setting current branch to checked out branch.
+# For example if this script is run on branch, test then,
+# commit will traverse from test branch of source repo to,
+# test branch of destination repo.
+CURRENT_BRANCH=`git rev-parse --abbrev-ref HEAD`
```
saurabhtripathics, line 21: I think we will need to change this after dev testing? Either we can pass this as parmeter to script, or provide a static value like develop.
Currently we are dynamically setting current branch to checked out branch.
For example if this script is run on branch, test than,
commit will traverse from test branch of source repo to,
test branch of destination repo.

acms_split.sh
```diff
@@ -0,0 +1,77 @@
...
+remote acquia_cms_search git@github.com:acquia/acquia_cms_search.git
+remote acquia_cms_starter git@github.com:acquia/acquia_cms_starter.git
+remote acquia_cms_support git@github.com:acquia/acquia_cms_support.git
+remote acquia_cms_toolbar git@github.com:acquia/acquia_cms_toolbar.git
+remote acquia_cms_tour git@github.com:acquia/acquia_cms_tour.git
+remote acquia_cms_video git@github.com:acquia/acquia_cms_video.git
+
+# Calling split method for mapping remote branches to splits.
+split 'modules/acquia_cms_article' acquia_cms_article
+split 'modules/acquia_cms_audio' acquia_cms_article
```
vishalkhode1, line 64: @saurabhtripathics All have same remote names i.e acquia_cms_article. Please update.

saurabhtripathics: done

acms_split.sh
```diff
@@ -0,0 +1,78 @@
...
+split 'modules/acquia_cms_page' acquia_cms_page
+split 'modules/acquia_cms_person' acquia_cms_person
+split 'modules/acquia_cms_place' acquia_cms_place
+split 'modules/acquia_cms_search' acquia_cms_search
+split 'modules/acquia_cms_starter' acquia_cms_starter
+split 'modules/acquia_cms_support' acquia_cms_support
+split 'modules/acquia_cms_toolbar' acquia_cms_toolbar
+split 'modules/acquia_cms_tour' acquia_cms_tour
+split 'modules/acquia_cms_video' acquia_cms_video
+
```
vishalkhode1, line 78: @chandan-singh7929 Add missing.
```suggestion
split 'modules/acquia_cms_document' acquia_cms_document
```

acms_split.sh
vishalkhode1: Yes, after dev testing we can change it and pass it as a parameter to script. If we don't, then that should be fine as well. We can checkout to develop/qa/stage branch and run this script. This will also do the same thing. :)

acms_split.sh
chandan-singh7929: @vishalkhode1 - Looks good to me.

vishalkhode1: Sorry, my bad. Above comment is for @saurabhtripathics. Please add the above missing code.

saurabhtripathics: done.

vishalkhode1 (approved):

kdruckman (approved):

PR #800, opened 2021-05-12
ACMS-763: Acquia CMS errors out in CloudIDE

saurabhtripathics:
https://backlog.acquia.com/browse/ACMS-763
**Motivation**
Acquia CMS errors out in CloudIDE
**Proposed changes**
Check that function exists in patches/d9-acms-settings.patch:

   // Only call this function on the cloud, not on a local environment.
    if (function_exists(acquia_hosting_db_choose_active) {
      acquia_hosting_db_choose_active();
    } 
 

**Testing steps**
 Following the installation steps for Acquia CMS from: https://docs.acquia.com/acquia-cms/install-acquia-cms/ -> https://github.com/acquia/acquia-cms-project should work in the Cloud IDE, and local.

### DIFF
--- a/patches/d9-acms-settings.patch
+++ b/patches/d9-acms-settings.patch
@@ -27,7 +27,10 @@ if (AcquiaDrupalEnvironmentDetector::isAhEnv()) {
       ],
     ];
     $databases = array_merge_recursive($databases, $default_settings);
-    acquia_hosting_db_choose_active();
+    // Only call this function on the cloud, not on a local environment.
+    if (function_exists(acquia_hosting_db_choose_active) {
+      acquia_hosting_db_choose_active();
+    }
   }
 }
 


marcelovani:
Looks good to me

vishalkhode1 (approved):

joshirohit100:
Used this patch on cloud IDE but failed as seems PHP syntax issue in patch.
This is missing closing parenthesis ")" for. 'function_exists'. Also the function name should be string. So this should be like -
```
if (function_exists('acquia_hosting_db_choose_active')) {
      acquia_hosting_db_choose_active();
    }
```

vishalkhode1:
Hi @joshirohit100 
Yes, It's fixed here: https://github.com/acquia/acquia_cms/pull/808/files

> Used this patch on cloud IDE but failed as seems PHP syntax issue in patch.
> This is missing closing parenthesis ")" for. 'function_exists'. Also the function name should be string. So this should be like -
> 
> ```
> if (function_exists('acquia_hosting_db_choose_active')) {
>       acquia_hosting_db_choose_active();
>     }
>

PR #803, opened 2021-05-12
ACMS-766: Update acquia_cms_starter module to support atomic nature

chandan-singh7929:
**Motivation**

Fixes #[ACMS-766](https://backlog.acquia.com/browse/ACMS-66)

**Proposed changes**

As we have to follow modular approach, where each acms module can be enable/disabled individually. Hence we have move configuration from profile install directory and leave at module level, so that if any module gets enable it will automatically enable its related configurations.

**Testing steps**

In order to test this, we need to run site install without acquia_cms_starter module once installed, enable module using drush & UI and check its working fine, you should be able to uninstall the module. 
 
**Merge requirements**
- [x] Manual testing by a reviewer


chandan-singh7929:
This would require #799 to be merge first

vishalkhode1 (approved):

PR #806, opened 2021-05-14
ACMS-746: Bundled content breaking config imports

chandan-singh7929:
**Motivation**

[Fixes #ACMS-746](https://backlog.acquia.com/browse/ACMS-746)

**Proposed changes**

Add configuration into ignore lists which are causing error when running blt setup second time
**Alternatives considered**

Delete default content for logo from acquia_cms_image moduel, along with ignore list for entity language
**Testing steps**

Follow steps mentioned in [BLT setup doc](https://docs.google.com/document/d/1jhk1aPp-ipOq0IXRRQ0xqhhcVPPtIJv4S_DtXahGvlY/edit#) use `composer require acquia/acquia_cms:dev-ACMS-746 --no-install` to test this PR
**Merge requirements**


modules/acquia_cms_common/src/Services/AcmsUtilityService.php
```diff
@@ -62,7 +62,10 @@ public function rebuildSiteStudio() {
     $batch = &batch_get();
     $batch = NULL;
     unset($batch);
-    return DX8CommandHelpers::rebuild([]);
+    return DX8CommandHelpers::rebuild([
+      'verbose' => '',
```
vishalkhode1, line 66: @chandan-singh7929 Don't we need to show verbose message anymore now ??

composer.json
```diff
@@ -34,7 +34,9 @@
         "drupal/autologout": "^1",
         "drupal/checklistapi": "^2.0",
         "drupal/collapsiblock": "^3",
+        "drupal/config_ignore": "^2.3",
         "drupal/config_rewrite": "1.3",
+        "drupal/config_split": "^1.7",
```
vishalkhode1, line 39: @chandan-singh7929 As I remember, you mentioned config_split is required by blt. Can you confirm, what issues you were facing, when we don't include this ?

modules/acquia_cms_common/src/Services/AcmsUtilityService.php
chandan-singh7929: I don't think we need to from here at least.

composer.json
chandan-singh7929: It will show error message like "config_split module not found" as error but will not hold the script execution.

vishalkhode1: @chandan-singh7929 I think we can skip this module and should not add in our project. Because this error is not due to our ACMS project, it's due to blt setup. So, In blt project itself, they can run the composer command to include this module and get rid of this error.

chandan-singh7929: make sense, let me update it.

chandan-singh7929: @vishalkhode1 - Its done, please have look.

vishalkhode1 (approved):

PR #808, opened 2021-05-17
ACMS-763: fix syntax error in patches file

saurabhtripathics:
Fix syntax error in patches file.

### DIFF
--- a/patches/d9-acms-settings.patch
+++ b/patches/d9-acms-settings.patch
@@ -28,7 +28,7 @@ if (AcquiaDrupalEnvironmentDetector::isAhEnv()) {
     ];
     $databases = array_merge_recursive($databases, $default_settings);
     // Only call this function on the cloud, not on a local environment.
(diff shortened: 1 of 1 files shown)


patches/d9-acms-settings.patch
```diff
@@ -28,7 +28,7 @@ if (AcquiaDrupalEnvironmentDetector::isAhEnv()) {
     ];
     $databases = array_merge_recursive($databases, $default_settings);
     // Only call this function on the cloud, not on a local environment.
-    if (function_exists(acquia_hosting_db_choose_active) {
+    if (function_exists(acquia_hosting_db_choose_active)) {
```
vishalkhode1, line 31: @saurabhtripathics I think it should be:
```suggestion
    if (function_exists('acquia_hosting_db_choose_active')) {
```
Ref: https://www.php.net/manual/en/function.function-exists.php

saurabhtripathics: done and tested on local @vishalkhode1

PR #815, opened 2021-05-20
ACMS-774: Acquia CMS Search test currently fails on develop

saurabhtripathics:
**Motivation**
Acquia CMS Search test currently fails on develop

**Proposed changes**

- Refactor phpunit test with respect to acquia search
- remove check related to solr search
- change some checks to conform with acquia search
- use session to fix field not found errors

**Testing steps**
Run test script to check, if test is passing

### DIFF
--- a/modules/acquia_cms_search/tests/src/Functional/AcquiaSearchFormIntegrationTest.php
+++ b/modules/acquia_cms_search/tests/src/Functional/AcquiaSearchFormIntegrationTest.php
@@ -52,40 +52,37 @@ class AcquiaSearchFormIntegrationTest extends BrowserTestBase {
    * Tests administrative integration with Acquia Search Solr.
    */
   public function testAcquiaSearchFormIntegration() {
-    $this->assertSame('database', Index::load('content')->getServerId());
-
-    $index = Index::load('acquia_search_index');
-    $this->assertTrue($index->status());
-    $this->assertSame('acquia_search_server', $index->getServerId());
-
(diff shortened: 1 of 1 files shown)


modules/acquia_cms_search/tests/src/Functional/AcquiaSearchFormIntegrationTest.php
```diff
@@ -52,40 +52,37 @@ class AcquiaSearchFormIntegrationTest extends BrowserTestBase {
...
-    // Our index should be using the Solr server, whereas the one that ships
-    // with Acquia Search Solr should be disabled, along with any views that are
-    // using it.
-    $this->assertSame('acquia_search_server', Index::load('content')->getServerId());
-    $index = Index::load('acquia_search_index');
-    $this->assertFalse($index->status());
-    $this->assertNull($index->getServerId());
-    $this->assertFalse(View::load('acquia_search')->status());
+    // Our index should be using the database server.
+    $this->assertSame('database', Index::load('content')->getServerId());
```
vishalkhode1, line 83: @saurabhtripathics Why did we change from acquia search to db search ?
Is this because this will set to acquia search only on Acquia environment ?

saurabhtripathics: yes @vishalkhode1 i checked on local, and this will be database only on local.

vishalkhode1 (approved):

PR #829, opened 2021-06-03
ACMS-788: Resolve issue to install & enable acquia_cms_page module independently.

vishalkhode1:
Resolved issue to install & enable acquia_cms_page module without need to install acquia_site_studio module.
The error was coming due to config: `field.storage.node.field_layout_canvas`. When we enable acquia_cms_page module, it tries to import this optional config, even though we have added `cohesion_elements` module as a dependency for this config.

When further debugged, I found that, when drupal tries to calculate missing dependencies for this config and it doesn't get `cohesion_elements` as a missing dependency, so drupal assume that, there's no missing dependency for this config and it try to import this config. So, during import it report the following error:
```Unable to determine class for field type 'cohesion_entity_reference_revisions' found in the 'field.storage.0' configuration ```

**Proposed changes**
Looked for the issue in drupal 9 core issue tracker and found that this issue was reported ([https://www.drupal.org/node/3180588](https://www.drupal.org/node/3180588)) and a patch is added for the same to resolve this.

**Testing steps**
Install the acms site and you should not see any error and the module acquia_cms_page should be enabled.

**Merge requirements**

### CODE REVIEW

**Comment by kdruckman** [comment]
I think this might depend on 790.

File: acquia_cms.info.yml [hunk 1]
```diff
@@ -6,7 +6,7 @@ distribution:
   install:
     finish_url: '/admin/tour'
 install:
-  - acquia_cms_common
+  - acquia_cms_page
```
chandan-singh7929, line 9: We should keep both `acquia_cms_common` & `acquia_cms_page`

chandan-singh7929:
I checked by installing site using `acquia_cms_page` as default in profile, installation completed without any error but when tried creating page content I don't see body field.
Screenshot 2021-06-04 at 3 40 40 PM

I think we should work on acms-790 first.

acquia_cms.info.yml
kdruckman: Common shouldn't be needed since it is required by the other component modules.

kdruckman (commented):
As above, I think this needs be incorporated in the work on 790 to be useful.

PR #841, opened 2021-06-15
ACMS-768: Added code for logging time for install rebuild start & end.

panshulK:
**Motivation**
* We need to calculate and send the install & rebuild start and end time to acquia telemetry
Fixes #ACMS-768

**Proposed changes**
* Add the install start time and end time and rebuild start time and end time and calculate the difference between the two & send to acquia telemetry via heartbeat_event in acquia_cms.profile

**Alternatives considered**
* None

**Testing steps**
<Pending>

**Merge requirements**

### DIFF
--- a/acquia_cms.profile
+++ b/acquia_cms.profile
@@ -12,6 +12,7 @@ use Drupal\acquia_cms\Form\SiteConfigureForm;
 use Drupal\cohesion\Controller\AdministrationController;
 use Drupal\cohesion_website_settings\Controller\WebsiteSettingsController;
 use Drupal\Core\Ajax\CloseDialogCommand;
+use Drupal\Core\Datetime\DrupalDateTime;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Installer\InstallerKernel;
 use Drupal\media_library\MediaLibraryState;
@@ -33,6 +34,46 @@ function acquia_cms_install_tasks_alter(array &$tasks) {
   // Decorate the site configuration form to allow the user to configure their
   // Cohesion keys at install time.
   $tasks['install_configure_form']['function'] = SiteConfigureForm::class;
+  // We want to capture the time when the installation starts.
+  // This code helps capture time right when the drupal bootstrap happens.
+  // The pre start function calls another method that performs the actual logic.
+  $tasks['install_bootstrap_full']['function'] = 'acquia_cms_pre_start';
+  $tasks['install_finished']['function'] = 'acquia_cms_install_ui_kit_finished';
+}
+
+/**
+ * Method that calls another method to capture the installation start time.
+ */
+function acquia_cms_pre_start($install_state) {
+  $function = $install_state['active_task'];
+  acquia_cms_set_install_time();
+  return $function($install_state);
+}
+
+/**
+ * Set the install start time using state API.
+ */
+function acquia_cms_set_install_time() {
+  $start_time = new DrupalDateTime();
+  $formatted_time = acquia_cms_format_time($start_time);
+  \Drupal::state()->set('install_start_time', $formatted_time);
+}
+
+/**
+ * Function for formatting date.
+ *
+ * @param \Drupal\Core\Datetime\DrupalDateTime $time
+ *   Parameter that passes raw date value.
+ *
+ * @return \Drupal\Core\Datetime\DrupalDateTime
+ *   The formatted date.
+ */
+function acquia_cms_format_time(DrupalDateTime $time) {
+  $formatted = \Drupal::service('date.formatter')->format(
+    $time->getTimestamp(), 'custom', 'Y-m-d h:i:s'
+  );
+
+  return $formatted;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


acquia_cms.profile
```diff
@@ -40,6 +40,8 @@ function acquia_cms_install_tasks_alter(array &$tasks) {
  */
 function acquia_cms_install_tasks(): array {
   $tasks = [];
+  $start_time = date('h:i:s a', time());
+  \Drupal::state()->set('install_start_time', $start_time);
```
vishalkhode1, line 44: @panshulK We can't add logic code directly here because this function gets invoked when each task runs.

acquia_cms.profile
```diff
@@ -88,16 +90,54 @@ function acquia_cms_install_tasks(): array {
  * Send heartbeat event after site installation.
  */
 function acquia_cms_send_heartbeat_event() {
+  $cohesion_configured = $config->get('api_key') && $config->get('organization_key');
+  // Get the install & rebuild time from state API.
+  $install_start_time = \Drupal::state()->get('install_start_time');
+  $install_end_time = \Drupal::state()->get('install_end_time');
+  $rebuild_start_time = \Drupal::state()->get('rebuild_start_time');
```
vishalkhode1, line 97: @panshulK It doesn't seems to be working correct, when site is installed using drush command. Because cohesion rebuild happen using `drush post:command` hooks.

acquia_cms.profile
```diff
@@ -40,6 +40,8 @@ function acquia_cms_install_tasks_alter(array &$tasks) {
  */
 function acquia_cms_install_tasks(): array {
   $tasks = [];
+  $start_time = date('h:i:s a', time());
```
vishalkhode1, line 43: @panshulK I think we need to include date month year as well. Because when you'll convert this time in unix timestamp `(strtotime())` this will assume date as today's date. So, this might result in incorrect time or negative time. 
Ex: Let's say we start installation today (i.e 18/06/2021) and install time starts at 11:45 pm and ends at 00:15 am i.e it took overall 30 mins. But when this time convert in unix timestamp, so actually it would convert following:
Start time: **18/06/2021 11:45 pm**. End time: **18/06/2021 00:15 am**, but actually it's incorrect as end time is: **19/06/2021 00:15 am**.

acquia_cms.profile
```diff
@@ -88,16 +90,54 @@ function acquia_cms_install_tasks(): array {
...
   \Drupal::service('acquia.telemetry')->sendTelemetry('acquia_cms_installed', [
     'Application UUID' => Environment::getAhApplicationUuid(),
     'Site Environment' => Environment::getAhEnv(),
+    'Install Timme' => $install_time_diff
+  ]);
+  \Drupal::service('acquia.telemetry')->sendTelemetry('acquia_cms_site_studio_rebuild_time', [
+    'Application UUID' => Environment::getAhApplicationUuid(),
+    'Site Environment' => Environment::getAhEnv(),
+    'Rebuild Time' => $rebuild_time_diff,
   ]);
```
vishalkhode1, line 115: @panshulK We need to send one more variable to telemetry i.e `acquia_cms_site_studio_installed`.

acquia_cms.profile
panshulK: Hey @vishalkhode1 
Moved this logic to the `pre_start` function.

acquia_cms.profile
panshulK: Can we discus more on this? I can see that the function used to set the start time and finished time is being called by both drush (cli) and non cli executions.

acquia_cms.profile
panshulK: Done @vishalkhode1

acquia_cms.profile
panshulK: Should we send this explicitly?
Because I can see that this is already being set and send as part of the code : 
```
  \Drupal::service('acquia.telemetry')->sendTelemetry('acquia_cms_installed', [
    'Application UUID' => Environment::getAhApplicationUuid(),
    'Site Environment' => Environment::getAhEnv(),
    'Install Timme' => $install_time_diff,
  ]);
  ```
  I just added the install_time_diff variable here.

Comment (vishalkhode1) [reply acquia_cms.profile:132]: @panshulK Yes we've to send this explicitly. With value true or false. 
True if site studio is installed, false if site studio is not installed.

Screenshot 2021-06-23 at 8 42 58 PM

Comment (panshulK) [reply acquia_cms.profile:159]: Done @vishalkhode1

**Review by vishalkhode1 — APPROVED** [review]

File: acquia_cms.profile [hunk 5]
```diff
@@ -33,6 +34,43 @@ function acquia_cms_install_tasks_alter(array &$tasks) {
   // Decorate the site configuration form to allow the user to configure their
   // Cohesion keys at install time.
   $tasks['install_configure_form']['function'] = SiteConfigureForm::class;
+  $tasks['install_bootstrap_full']['function'] = 'acquia_cms_pre_start';
+  $tasks['install_finished']['function'] = 'acquia_cms_install_ui_kit_finished';
+}
+
+/**
+ * Print acquia cms icon on terminal and then start first active install task.
```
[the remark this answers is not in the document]
panshulK: Done

acquia_cms.profile
```diff
@@ -33,6 +34,43 @@ function acquia_cms_install_tasks_alter(array &$tasks) {
   // Decorate the site configuration form to allow the user to configure their
   // Cohesion keys at install time.
   $tasks['install_configure_form']['function'] = SiteConfigureForm::class;
+  $tasks['install_bootstrap_full']['function'] = 'acquia_cms_pre_start';
```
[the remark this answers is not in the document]
panshulK: Done

PR #847, opened 2021-06-22
ACMS-786: Add samlauth config to fix failing integration tests.

kdruckman:
**Motivation**
Fixes ACMS-786

**Proposed changes**
Add samlauth configuration to failing tests that create users.

**Alternatives considered**
None.

**Testing steps**
Verify that integrated tests on current Drupal versions pass.

**Merge requirements**


modules/acquia_cms_common/tests/src/ExistingSite/ContentTypeListTestBase.php
```diff
@@ -40,6 +41,21 @@ abstract class ContentTypeListTestBase extends ExistingSiteBase {
   protected function setUp(): void {
     parent::setUp();
 
+    // If the samlauth module is installed, ensure that it is configured (in
+    // this case, using its own test data) to avoid errors when creating user
+    // accounts in this test.
+    if ($this->container->get('module_handler')->moduleExists('samlauth')) {
```
vishalkhode1, line 47: @kdruckman I think we can make following changes:

1.  We can add this code either as a function in trait and then call that trait & function across all files to avoid code duplicating or other thing we could do is: 
2. Create some base class like `AcquiaCmsExisingSiteBase` which would extends ExistingSiteBase class and add this code there and our all tests would then extends `AcquiaCmsExisingSiteBase` class.

vishalkhode1: Merging it for now.

PR #849, opened 2021-06-24
ACMS-794: Updated example profile to use claro theme instead of acquia_claro.

vishalkhode1:
**Motivation**

ACMS-794: Updated example profile to use claro theme instead of acquia_claro.


chandan-singh7929 (approved):
@vishalkhode1  - This looks good to me, I have added some suggestion on ticket please have a look and let me know if that make sense.
cc: @kdruckman

PR #853, opened 2021-06-28
ACMS-817: Update test dependencies in refactored modules.

kdruckman:
**Motivation**
Fixes ACMS-817, Ensure test dependencies are met in component modules.

**Proposed changes**
Add workbench_email and entity_clone as dependencies in acquia_cms_common, remove cohesion dependency in component module tests and filter_html format.

**Alternatives considered**

**Testing steps**
When applied, these changes should allow all component module tests to pass when run in isolation.

**Merge requirements**


chandan-singh7929:
@kdruckman - I tested this PR and all test mentioned in ticket ACMS-817 are passing, hence making this available for review.

modules/acquia_cms_site_studio/acquia_cms_site_studio.install
```diff
@@ -64,3 +64,19 @@ function set_theme_site_studio_as_default() {
...
+    ->getEditable('filter.format.filtered_html')
+    ->set('filters.black_list_html_tags', [
+      'id' => 'black_list_html_tags',
+      'provider' => 'cohesion',
+      'status' => 'false',
+      'weight' => '-41',
+      'settings' => '',
+    ])
+    ->save();
+}
```
chandan-singh7929, line 82: @kdruckman - I see function `set_cohesion_filter()` call from anywhere.

modules/acquia_cms_common/acquia_cms_common.info.yml
```diff
@@ -7,6 +7,7 @@ dependencies:
   - drupal:ckeditor
   - drupal:content_moderation
   - drupal:media
+  - drupal:media_library
```
kdruckman, line 10: @vishalkhode1 @chandan-singh7929 Do you see any reason not to require media_library here? In order to set the filtered_html format config when the site studio module is installed, the filter.format.filtered_html.yml from the common module needs to be in place, which means it shouldn't be in optional config, but since there is a dependency on some media_library config in the filtered_html config, this module would need to require it. Normally, when the full profile is installed, place requires media_library, so that dependency is satisfied. Does this approach makes sense to you? The other option might be to dynamically create it on install of the common module if we don't want to add dependencies.

chandan-singh7929: This make sense to me, currently we have added dependency where it is require ex:
- acquia_cms_place.info.yml
- acquia_cms_place.info.yml

Maybe we can remove from above if we are going to provide from common, since they already require common as dependency.

vishalkhode1 (approved):

composer.json
```diff
@@ -225,6 +225,7 @@
         "build:local": [
             "mkdir -p $PROFILE_DIR $MODULES_DIR",
             "find $PWD -name 'acquia_cms.*' -type f -maxdepth 1 -exec ln -s -f '{}' $PROFILE_DIR ';'",
+            "find ./docroot/modules/contrib -type d -name \"acquia_cms*\" -exec mv '{}' ./docroot/modules ';'",
```
vishalkhode1, line 228: @chandan-singh7929 Can you check, if this change is from your ticket ?

acms_split.sh
```diff
@@ -53,6 +53,7 @@ remote acquia_cms_page git@github.com:acquia/acquia_cms_page.git
 remote acquia_cms_person git@github.com:acquia/acquia_cms_person.git
 remote acquia_cms_place git@github.com:acquia/acquia_cms_place.git
 remote acquia_cms_search git@github.com:acquia/acquia_cms_search.git
+remote acquia_cms_site_studio git@github.com:acquia/acquia_cms_site_studio.git
```
vishalkhode1, line 56: @chandan-singh7929 Can you confirm if this change is from your ticket ?

modules/acquia_cms_common/src/Commands/AcmsConfigImportCommands.php
```diff
@@ -567,8 +570,8 @@ public function validateConfigResetCommand(CommandData $commandData) {
     }
     // Get scope from user input.
     if ($isInteractive && empty($messages) && !$scope) {
-      $scope = $this->io()->choice(dt('Choose a scope.'), self::ALLOWED_SCOPE, NULL);
-      $commandData->input()->setOption('scope', self::ALLOWED_SCOPE[$scope]);
+      $scope = $this->io()->choice(dt('Choose a scope.'), $this->allowedScope, NULL);
```
vishalkhode1, line 573: @chandan-singh7929 Why this changes is showing in your PR ? I think it's not from your ticket,

vishalkhode1 (changes requested):

vishalkhode1:
@chandan-singh7929 I think, it was probably due to conflicts. Ignore this, I have resolved the conflicts now.

vishalkhode1 (approved):

PR #868, opened 2021-07-15
ACMS-768: Code refactored for sending Site Studio profiling metrics to Telemetry.

vishalkhode1:
ACMS-768: Code refactored for sending Site Studio profiling metrics to Telemetry.

**Motivation**

Fixes Timezone issue & refactored code to implement it using OOPS.

**Testing steps**
- Install Site using UI and on Site Configuration page, change Country and Timezone.

**Merge requirements**
- [ ] _Major change_

### DIFF
--- a/acquia_cms.profile
+++ b/acquia_cms.profile
@@ -6,13 +6,13 @@
  */
 
 use Acquia\DrupalEnvironmentDetector\AcquiaDrupalEnvironmentDetector as Environment;
+use Acquia\Utility\AcquiaTelemetry;
 use Drupal\acquia_cms\Facade\CohesionFacade;
 use Drupal\acquia_cms\Facade\TelemetryFacade;
 use Drupal\acquia_cms\Form\SiteConfigureForm;
 use Drupal\cohesion\Controller\AdministrationController;
 use Drupal\cohesion_website_settings\Controller\WebsiteSettingsController;
 use Drupal\Core\Ajax\CloseDialogCommand;
-use Drupal\Core\Datetime\DrupalDateTime;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Installer\InstallerKernel;
 use Drupal\media_library\MediaLibraryState;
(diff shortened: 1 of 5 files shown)


tests/src/Kernel/AcquiaTelemetryTest.php
```diff
@@ -0,0 +1,78 @@
...
+    $this->telemetry->setTime('start_time', "2021-07-15 12:00:00 +05:30");
+    $this->telemetry->setTime('end_time', "2021-07-15 08:00:00 +01:00");
+    $this->assertSame(1800, $this->telemetry->calculateTime('start_time', 'end_time'));
+    $this->clearValues(['start_time', 'end_time']);
+  }
+
+  /**
+   * Should throw exception when time difference is negative.
+   */
+  public function testNegativeTimeExpectException() {
```
chandan-singh7929, line 53: @vishalkhode1 - This look pretty clean, just one suggestion, instead of testing exception shouldn't we simply check for positive value?

chandan-singh7929 (approved):

acquia_cms.profile
```diff
@@ -382,9 +302,8 @@ function acquia_cms_rebuild_site_studio() {
 function acquia_cms_rebuild_site_studio_finished() {
   // The 'success' parameter means no fatal PHP errors were detected. All
   // other error management should be handled using 'results'.
```
panshulK, line 304: @vishalkhode1 Can we alter the comments here a bit? Are these comments related to Telemtery class itself?

panshulK (approved):
@vishalkhode1 

Changes looks good to me. Only a minor comment related change is required.

tests/src/Kernel/AcquiaTelemetryTest.php
vishalkhode1: Yes, we did check for positive value as well as for the negative value.

acquia_cms.profile
vishalkhode1: No, that's a general comment on how batch API works. That is, it returns the results array. If array contains key: `error` means, batch doesn't executed successfully and it reported some error.

vishalkhode1: But yes, ideally it shouldn't be there.

PR #875, opened 2021-07-23
ACMS-834: Created patch in Site Studio module to fix Access denied error.

vishalkhode1:
**Motivation**

Fixes access denied issue when cohesion module is installed during site installation.

**Proposed changes**
Created a patch in Site Studio (cohesion) module which causing the access denied error.

**Alternatives considered**
Alternate patch is tried as provided by Site Studio team (i.e https://backlog.acquia.com/secure/attachment/323751/323751_ACO-1169.patch) but it didn't solve the actual issue.

**Testing steps**
- Install the Site using UI (without adding Site Studio keys in Site Configuration page). After site installation is completed, user shouldn't be redirected to Access Denied page.
- Install the Site using UI (adding Site Studio keys in Site Configuration page). After site installation is completed, user shouldn't be redirected to Access Denied page.

**Merge requirements**
- [ ] _Bug_, _Major_

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -153,6 +153,9 @@
             "drupal/core": "-p2"
         },
         "patches": {
+            "acquia/cohesion": {
+                "Site install using UI gives 403 on finish, due to update core 9.2": "https://gist.githubusercontent.com/vishalkhode1/8762f8e2e5569f01a3fdcf3b8d8e2b50/raw/3f916d17b717fd5f3c215bde02e741a9ee7be4e5/site-studio-access-denied.patch"
+            },
             "drupal/acquia_telemetry-acquia_telemetry": {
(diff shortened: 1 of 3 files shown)


kdruckman:
@vishalkhode1 This looks great to me, and nice detective work. Looking at the changes in SharedTempStoreFactory::get from D9.1 to 9.2, your approach of passing the session id makes sense to me. I tested this by manually installing with and without Site Studio credentials and both worked as expected.

vishalkhode1:
> @vishalkhode1 This looks great to me, and nice detective work. Looking at the changes in SharedTempStoreFactory::get from D9.1 to 9.2, your approach of passing the session id makes sense to me. I tested this by manually installing with and without Site Studio credentials and both worked as expected.

Excellent, then merging it. So we can dev/QA test and include in our v1.2.5 release. 
Thanks.

PR #933, opened 2021-09-08
ACMS-901: Updated the Modules Using composer.

panshulK:
**Motivation**
* We need to update the Contributed module to required version.
Fixes #NNN

**Proposed changes**
* Update module using composer

**Alternatives considered**
* None

**Testing steps**
* Update modules and test the changes accordingly.
* Mark the progress of the modules updated here - https://docs.google.com/spreadsheets/d/1Aki3M1n72YLPEGZUWwKwHXLckJiq8z8W6cwNMy4QfZw/edit#gid=0

**Merge requirements**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,14 +39,14 @@
         "drupal/core": "~9.1.12 || ^9.2.4",
         "drupal/default_content": "^2",
         "drupal/diff": "^1",
-        "drupal/entity_clone": "1.0-beta4",
+        "drupal/entity_clone": "1.0-beta6",
         "drupal/facets": "^1.8",
         "drupal/facets_pretty_paths": "^1.1",
         "drupal/field_group": "^3",
         "drupal/focal_point": "1.5",
         "drupal/geocoder": "3.20",
         "drupal/geofield": "^1",
-        "drupal/google_analytics": "^2",
+        "drupal/google_analytics": "^3.1",
         "drupal/google_tag": "^1",
         "drupal/honeypot": "^2.0",
         "drupal/imagemagick": "^3",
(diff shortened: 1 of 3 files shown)


panshulK:
Google Analytics Module requires Drush update to be executed.

vishalkhode1 (approved):

composer.json
```diff
@@ -39,14 +39,14 @@
         "drupal/core": "~9.1.12 || ^9.2.4",
         "drupal/default_content": "^2",
         "drupal/diff": "^1",
-        "drupal/entity_clone": "1.0-beta4",
+        "drupal/entity_clone": "^1.0@beta",
```
vishalkhode1, line 42: @panshulK Update the module's composer.json file as well. Ex: `modules/acquia_cms_common/composer.json`
Same for all the below modules/patches which are updated.

chandan-singh7929: Also, If we are patching any module we should pin the module version for which patch has been give otherwise that patch will start failing if something changed in the next release of the module.

chandan-singh7929 (approved):
Since we are updating modules, we should check the same with custom profile in isolation as well to make sure dependencies are also updated.

composer.json
panshulK: Done @chandan-singh7929.

@vishalkhode1  Updated and checked the composer.json file for all modules. The updates were only required to be made to acquia_cms_common module's composer json file.

panshulK:
> Overall changes looks good to me.
> Since we are updating modules, we should check the same with custom profile in isolation as well to make sure dependencies are also updated.

Done @chandan-singh7929.

PR #941, opened 2021-09-21
ACMS-839: Changed the email ID to example no reply.

panshulK:
**Motivation**
* The email id configured at various places is no reply acquia. It should be changed to no reply example.
Fixes #NNN

**Proposed changes**
* Change email ID to no reply example.

**Alternatives considered**
* None

**Testing steps**
* Check that all places have new email ID configured and the mail is only sent if the mail id is changed.

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_common/acquia_cms_common.module
+++ b/modules/acquia_cms_common/acquia_cms_common.module
@@ -220,3 +220,14 @@ function _acquia_cms_common_update_view_display_options_style($view_name, $displ
     $view_storage->save($view);
   }
 }
+
+/**
+ * Implements hook_mail_alter().
+ */
+function acquia_cms_common_mail_alter(&$message) {
+  if (strpos($message['key'], 'template::') === 0) {
+    if ((!empty($message['from'])) && ($message['from'] === 'no-reply@example.com')) {
+      $message['send'] = 'false';
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


modules/acquia_cms_common/acquia_cms_common.module
```diff
@@ -220,3 +221,17 @@ function _acquia_cms_common_update_view_display_options_style($view_name, $displ
     $view_storage->save($view);
   }
 }
+
+/**
+ * Implements hook_mail_alter().
+ */
+function acquia_cms_common_mail_alter(&$message) {
+  if (strpos($message['key'], 'template::') === 0) {
+    if (AcquiaDrupalEnvironmentDetector::isLocalEnv()) {
```
vishalkhode1, line 230: @panshulK I think we can't use this, as this assumes everything as local if it's not an Acquia environment. So, it would restrict sending email, if acquia_cms is used in any other hosting platform.
@kdruckman

vishalkhode1: Please also see if we can add some test coverage around this functionality.

panshulK: Sure @vishalkhode1 
Can we take the test creation part separately as this might need some more discussion?

Thanks

vishalkhode1 (approved):

PR #948, opened 2021-09-28
PF-3770: revising package load order.

mikemadison13:
**Motivation**
Packages are not loading in the proper order during install, causing site-install to fail.

Fixes #PF-3770

**Proposed changes**
- combine the base unit and responsive grid settings into a single file
- import this file first
- prioritize fonts during import to prevent subsequent failure


mikemadison13:
something is up with the patch. the fix works if i make it manually, but the patch isn't actually renaming the files as i want it to (so it's still failing).

chandan-singh7929:
Test are failing due to missing following variables:
```
export CONNECTOR_ID=<CONNECTOR_ID>
export SEARCH_UUID=<SEARCH_UUID>
export CONNECTOR_KEY=<CONNECTOR_KEY>
```
I have create one in ACMS #962 by cloning this so that can make any changes if required.

PR #954, opened 2021-10-04
ACMS-924: Changed the template to View template.

panshulK:
**Motivation**
* The article slider is not appearing correctly as per the design.
Fixes #NNN

**Proposed changes**
* Changed the template to view template for article slider.

**Alternatives considered**
* None

**Testing steps**
* Once changes are deployed, create a content of type page and add article slider component to the page.
* Save the changes in layout canvas and on page display check that the slider is appearing correctly as per the design.

**Merge requirements**
Screenshot 2021-10-04 at 3 41 17 PM

### DIFF
--- a/modules/acquia_cms_common/src/EventSubscriber/ConfigEventsSubscriber.php
+++ b/modules/acquia_cms_common/src/EventSubscriber/ConfigEventsSubscriber.php
@@ -64,7 +64,7 @@ class ConfigEventsSubscriber implements EventSubscriberInterface {
           break;
 
         case 'views.view.article_cards':
-          _acquia_cms_common_update_view_display_options_style('article_cards', 'recent_articles_block', 'view_tpl_article_cards_slider');
+          _acquia_cms_common_update_view_display_options_style('article_cards', 'default', 'view_tpl_article_cards_slider');
           break;
       }
     }


modules/acquia_cms_article/config/optional/views.view.article_cards.yml
```diff
@@ -52,7 +53,10 @@ display:
           items_per_page: 7
           offset: 0
       style:
-        type: default
+        type: cohesion_layout
+        options:
+          views_template: view_tpl_article_cards_slider
+          master_template: __none__
```
chandan-singh7929, line 59: @panshulK - We can't update views template directly here, this will throw error when site install with standard profile or where site studio is not preset, instead we can update this configuration based on modules availability like we have done for other configuration using EventSubscriber. 
You can take reference from here: https://github.com/acquia/acquia_cms/blob/develop/modules/acquia_cms_common/src/EventSubscriber/ConfigEventsSubscriber.php#L54-L70
cc: @vishalkhode1

panshulK: Thanks for sharing this @chandan-singh7929 , checking!

panshulK: Hey @chandan-singh7929 
Reverted the changes. Please review!

Thanks

vishalkhode1 (approved):

PR #957, opened 2021-10-05
ACMS-924: Added update hook for updating the view

panshulK:
**Motivation**
* The article slider is not appearing correctly as per the design.
Fixes #NNN

**Proposed changes**
* Changed the template to view template for article slider.

**Alternatives considered**
* None

**Testing steps**
* Once changes are deployed, create a content of type page and add article slider component to the page.
* Save the changes in layout canvas and on page display check that the slider is appearing correctly as per the design.

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_article/acquia_cms_article.install
+++ b/modules/acquia_cms_article/acquia_cms_article.install
@@ -22,9 +22,14 @@ function acquia_cms_article_install() {
 
 /**
  * Implements hook_update_N().
+ *
+ * Forcefully import the Article Cards View configuration
+ * so that the EventSubscriber responsible for updating the
+ * view template format can act accordingly.
  */
 function acquia_cms_article_update_8001() {
-  if (\Drupal::moduleHandler()->moduleExists('acquia_cms_common') && \Drupal::moduleHandler()->moduleExists('acquia_cms_site_studio')) {
-    _acquia_cms_common_update_view_display_options_style('article_cards', 'default', 'view_tpl_article_cards_slider');
+  $config = \Drupal::configFactory()->getEditable('views.view.article_cards');
+  if (!empty($config)) {
+    $config->save();
   }
 }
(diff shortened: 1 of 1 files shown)


vishalkhode1 (approved):
Looks good to me. @chandan-singh7929 can you also check this ?

modules/acquia_cms_article/acquia_cms_article.install
```diff
@@ -19,3 +19,12 @@ function acquia_cms_article_install() {
     'delete any article content',
   ]);
 }
+
+/**
+ * Implements hook_update_N().
```
chandan-singh7929, line 24: Instead of writing hook_update_N, can we add some details what this update is doing here?

modules/acquia_cms_article/acquia_cms_article.install
```diff
@@ -19,3 +19,12 @@ function acquia_cms_article_install() {
...
+ */
+function acquia_cms_article_update_8001() {
+  if (\Drupal::moduleHandler()->moduleExists('acquia_cms_common') && \Drupal::moduleHandler()->moduleExists('acquia_cms_site_studio')) {
+    _acquia_cms_common_update_view_display_options_style('article_cards', 'default', 'view_tpl_article_cards_slider');
+  }
```
chandan-singh7929, line 29: Also If I have to update this I would simply import the configuration again  and rest will be taken care by the eventSubscriber, this hook should be implemented as part of acqua_cms_article module itself instead of common let me know if you have any queries.

chandan-singh7929 (changes requested):
I think this need some work

modules/acquia_cms_article/acquia_cms_article.install
panshulK: Done

modules/acquia_cms_article/acquia_cms_article.install
panshulK: Done @chandan-singh7929 
The Update hook is already in acquia_cms_article but updated the code to save view file programatically in update hook.

chandan-singh7929 (approved):
LGT

PR #987, opened 2021-11-15
ACMS-960: Added code to save the MAPS key from environment variables.

panshulK:
**Motivation**
* The API key is not set until and unless it is set for Cohesion key settings.
Fixes #NNN

**Proposed changes**
* Once the cohesion keys are set the jobs will not fail.

**Alternatives considered**
* None

**Testing steps**
* Install site and verify that maps are working correctly if ENV variable for MAPS API is set.

**Merge requirements**


modules/acquia_cms_site_studio/acquia_cms_site_studio.install
```diff
@@ -56,6 +56,18 @@ function acquia_cms_site_studio_install() {
         ->save();
     }
   }
+  // Starter "really" needs Google Maps API Keys to "work."
+  // This code is a work around for the dev team to ensure that we can install
+  // these modules w/o issue until a more permanent API key is supplied by
+  // Acquia AND/OR we provide a free maps option.
+  // @todo remove this once the above conditions are met.
+  $gmaps_key = getenv('GMAPS_KEY');
+  if (isset($gmaps_key)) {
```
vishalkhode1, line 65: @panshulK This should be `!empty` instead of `isset` and this can go inside function:
`function set_site_studio_credentials` Also, comment should simply something like: Set the google map, if it's present in environment variable. (As starter module is something we use internally and I don't think we've to add comment in our codebase about it).

vishalkhode1: @panshulK Also we've to commit articats file as well for it.

PR #992, opened 2021-11-22
DIT-445: add gitignore and gitattributes files

amarlata-tect:


vishalkhode1:
Closing this for now. Will create a ticket in ACMS to fix this.

PR #994, opened 2021-12-01
ACMS-972: Add a param to install command to import from custom config

chandan-singh7929:
**Motivation**

Fixes #[ACMS-972](https://backlog.acquia.com/browse/ACMS-972)

**Proposed changes**

Improve installation time with and without existing configuration.

**Alternatives considered**

NA

**Testing steps**

- Install the site with acquia_cms and observe the installation time taken.

**Merge requirements**

### DIFF
--- a/acquia_cms.profile
+++ b/acquia_cms.profile
@@ -212,22 +212,6 @@ function acquia_cms_modules_uninstalled(array $modules) {
 }
 
 /**
- * Implements hook_module_implements_alter().
- */
-function acquia_cms_module_implements_alter(array &$implementations, string $hook) : void {
-  // @todo The below code needs to be updated once the memory limit issue is
-  // fixed by the site studio.
-  if ($hook === 'modules_installed') {
-    // Prevent cohesion_sync from reacting to module installation, for an
-    // excellent reason: it tries to import all of the new module's sync
-    // packages, at once, in the current request, which leads to memory errors.
-    // We replace it with a slightly smarter implementation that uses the batch
-    // system when installing a module via the UI.
-    unset($implementations['cohesion_sync']);
-  }
-}
-
-/**
  * Method that calls another method to capture the installation end time.
  */
 function install_acms_finished() {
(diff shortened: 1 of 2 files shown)


vishalkhode1:
@chandan-singh7929 I think we now don't need following functions:
1. `acquia_cms_module_implements_alter()`
2. `acquia_cms_modules_installed()`

If not, then we get them removed ?

chandan-singh7929:
> 2\. `acquia_cms_modules_installed()`

@vishalkhode1 - I think above one is required, if module get enable individually and have some site studio components then it should import. Though we can remove implements alter as we have same code in common module.

vishalkhode1:
> > 2. `acquia_cms_modules_installed()`
> 
> @vishalkhode1 - I think above one is required, if module get enable individually and have some site studio components then it should import. Though we can remove implements alter as we have same code in common module.

That anyway will install. Right ?
Because we've unset implementations only when site is getting installed ?
```
if ($hook == 'modules_installed' && InstallerKernel::installationAttempted()) {
  unset($implementations['cohesion_sync']);
}

chandan-singh7929:
@vishalkhode1 - Changes done.

vishalkhode1:
> > > 2. `acquia_cms_modules_installed()`
> > 
> > 
> > @vishalkhode1 - I think above one is required, if module get enable individually and have some site studio components then it should import. Though we can remove implements alter as we have same code in common module.
> 
> That anyway will install. Right ? Because we've unset implementations only when site is getting installed ?
> 
> ```
> if ($hook == 'modules_installed' && InstallerKernel::installationAttempted()) {
>   unset($implementations['cohesion_sync']);
> }
> ```
As discussed, we'll remove `acquia_cms_module_implements_alter()` as we've moved the code to `acquia_cms_common.module` file and we'll keep function `acquia_cms_modules_installed()` for now. If needed, we'll create a separate ticket to verify and then remove.

PR #995, opened 2021-12-02
ACMS-976: Fixed deprecated code warnings: Call to deprecated method getSession().

vishalkhode1:
**Motivation**
Fixes `STRICT_DEPRECATED_CODE_SCAN` travis job.

**Proposed changes**
The method `$element->getSession()` has been marked as deprecated from mink: 1.6. See [here](https://github.com/minkphp/Mink/blob/master/CHANGES.md#160--2014-09-26). 
In class CohesionElement.php (tests/src/ExistingSiteJavascript/CohesionElement.php) we've used `$element->getSession()` & `$this->getSession()` at multiple places. But since the method `getSession()` from element is marked as deprecated, so it's giving error. To fix this, we've added a protected variable `$session` in the class and use that instead everywhere (where session is required)

**Alternatives considered**
N/A

**Testing steps**
In order to test in local, we need drupal-check command. Use the following commands to download drupal-check command and test:
1. composer require mglaman/drupal-check --dev
2. ./vendor/bin/drupal-check tests/src --memory-limit=-1

You shouldn't see any error & warnings when running above drupal-check command.
Also, run the complete site tests to make sure all tests are passing i.e `./acms-run-tests.sh `

**Merge requirements**
- [ ] _Major change_ applied

### DIFF
--- a/tests/src/ExistingSiteJavascript/CohesionElement.php
+++ b/tests/src/ExistingSiteJavascript/CohesionElement.php
@@ -4,6 +4,7 @@ namespace Drupal\Tests\acquia_cms\ExistingSiteJavascript;
 
 use Behat\Mink\Element\ElementInterface;
 use Behat\Mink\Element\NodeElement;
+use Behat\Mink\Session;
 use Drupal\Tests\acquia_cms\Traits\AwaitTrait;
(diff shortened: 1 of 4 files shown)


chandan-singh7929 (approved):

PR #1000, opened 2021-12-09
ACMS-979: Fixed the patch failing issue on Drupal 9.3

vishalkhode1:
**Motivation**

Fixes the patch failing issue on Drupal 9.3

**Proposed changes**
- Removed the patch and added hook_system_breadcrumb_alter() to set breadcrumb texts for node create/edit page (if it's empty)

**Alternatives considered**
Tried to use following composer plugin: https://github.com/netresearch/composer-patches-plugin but it's not working and giving error. It looks like it's conflicting with the composer plugin which we are using currently 
i.e `cweagans/composer-patches`

**Testing steps**
- Go to any content type create page and there you should see breadcrumb as: `Home -> Node -> Add content`.
- Go to any node edit page and there you should see breadcrumb as: `Home -> <Node Title>`.
- Navigate to Frontpage view (i.e /admin/structure/views/view/frontpage) & change to display title to anything like: 
`Some Title` and now navigate again to node create/edit page and you should see breadcrumb starting as: 
`Home -> Some Title`

**Merge requirements**
- [ ] _Major change_,
- [ ] Manual testing by a reviewer
- [ ] PHPUnit FunctionalTest added.

PR #1002, opened 2021-12-09
Update patch from #3220437 to apply to 9.3.x with no other changes

balsama:
**Motivation**
ORCA builds are broken because the existing patch doesn't apply to 9.3.x.

**Proposed changes**
Same exact end user behavior and contents of the patch. The patched file had some keys reordered so the patch needed to be updated.

**Alternatives considered**
There is an alternative (better?) approach in https://github.com/acquia/acquia_cms/pull/1000. But that work wasn't completed prior to 9.3.0 being released. This is intended as a hotfix to unblock the rest of engineering.

**Merge requirements**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -163,7 +163,7 @@
                 "Support for export filtering via Drush": "https://www.drupal.org/files/issues/2021-08-18/config_ignore_2857247-75.patch"
             },
             "drupal/core": {
-                "Frontpage View Title for breadcrumb trail": "https://www.drupal.org/files/issues/2021-07-01/core_frontpage_views_title_patch.patch",
+                "3220437 - Empty breadcrumb at node/add and node/add/{content_type} when frontpage view is enabled\nl": "https://www.drupal.org/files/issues/2021-12-09/3220437-69--do-not-test.txt",
                 "It is possible to overflow the number of items allowed in Media Library": "https://www.drupal.org/files/issues/2019-12-28/3082690-80.patch",
                 "Media Library widget produces error this value should not be null when field is required": "https://www.drupal.org/files/issues/2020-10-08/3160238-25.patch",
                 "SQLite database locking errors cause fatal errors": "https://www.drupal.org/files/issues/1120020-59.patch"


effulgentsia:
Does this apply to 9.2 though? I think the issue is ACMS needing to support 9.2 and 9.3 and wanting to do so with the same codebase (i.e., not branch).

balsama:
> Does this apply to 9.2
It does not. Or at least I suspect it does not.

In my opinion, ACMS does needs a workflow to follow when that's the case though, and I'd be surprised if they didn't. In similar projects that need to support two (or more) minor/major versions at the same time, there is usually a policy to branch the project too.

E.g., for Lightning, if a component _did_ work with more than one branch at a time, we would (celebrate and) continue development there with a loose core version constraint. When a patch failed to apply, we would usually create a new major of the component and tighten up the constraints. I suspect this is even easier now that contrib projects have full semver support.

Of course, if ACMS _can't_ do this for some reason, this PR won't work.

effulgentsia:
For this specific case, #1000 gets around it by altering the config in a Drupal hook rather than via a patch.

For the case of managing patches in general, @danepowell made a good suggestion in https://github.com/cweagans/composer-patches/issues/351#issuecomment-990060308 that a large project (like ACMS) can put all of its patches in a subpackage, so that only that subpackage would need to have multiple branches. Lightning did something like this by having all of Lightning's patches to Drupal core in a lightning_core module, and thereby Lightning's other components and the distribution as a whole didn't need separate branches for different Drupal core versions.

kdruckman:
FWIW, I assumed ACMS would eventually need multiple branches, just as Lightning had 5.0.x for 9.0, 5.1.x for 9.1.x, etc, but if Dane's suggested approach is doable, that might ease the maintenance burden.

balsama:
> For this specific case, #1000 gets around it by altering the config in a Drupal hook rather than via a patch

100%. I even mentioned that #1000 is likely a "better" approach. But its tests are failing (albeit just like this one's). And it's not merged. Since so many people were blocked on this today, I was just trying to provide an alternative in case that approach is somehow blocked.

balsama:
Looks like #1000 was merged (🎉) which makes this obsolete.

PR #1004, opened 2021-12-13
ACMS-970: Added code to programmatically create facet.

panshulK:
**Motivation**
* The search module requires the Article content type to be enabled and article enables other content types. The search module should be enabled in isolation.
Fixes #NNN

**Proposed changes**
* Created the search_category facet programmatically because facet has the dependency on Content types. So the facet gets created only when the content type modules are enabled (the content types are created).

**Alternatives considered**
* None

**Testing steps**
* Scenario 1 (Custom Profile and Enabling the module in isolation)-
** Tested the module in isolation (Using a custom profile and enabling only the search module)
** Visit the /search page once the module is enabled and notice that the /search page opens without any errors - If there is not content created in the system the search page would show up empty.
** Enable any content type module example "acquia_cms_person" and notice that the facet `search_category` is created at page - /admin/config/search/facets/search_category/edit (login as admin user)

* Scenario 2 (Acquia CMS installation)-
** Tested the working of the module in acquia_cms.
** Run simple acquia_cms installation with SS keys and enable the `acquia_cms_starter` module.
** Visit the /search page and notice the search results and all the facets are working correctly.

**Merge requirements**


modules/acquia_cms_search/acquia_cms_search.module
```diff
@@ -161,3 +162,80 @@ function acquia_cms_search_search_api_server_update(EntityInterface $entity) {
 function acquia_cms_search_form_acquia_cms_solr_search_form_alter(array &$form) {
   $form['#submit'][] = AcquiaSearchFacade::class . '::submitSettingsForm';
 }
+
+/**
+ * Implements hook_modules_installed().
+ */
+function acquia_cms_search_modules_installed(array $modules) {
+  $module_handler = Drupal::moduleHandler();
```
vishalkhode1, line 170: @panshulK I think we do not need to use module_handler service as `$array` contains the modules which were installed.
We can use that. Ex:
```suggestion
$acmsModules = ["acquia_cms_person", "acquia_cms_place", "acquia_cms_article", "acquia_cms_event", "acquia_cms_page"];
if (array_intersect($acmsModules, $modules)) {
  // Create facets.
}
```

modules/acquia_cms_search/acquia_cms_search.module
```diff
@@ -161,3 +162,80 @@ function acquia_cms_search_search_api_server_update(EntityInterface $entity) {
...
+        ],
+        'show_reset_links' => 'false',
+        'reset_text' => 'Show all',
+        'hide_reset_when_no_selection' => 0,
+      ]);
+      $facet->setWeight(0);
+
+      // Set default empty behaviour & other settings.
+      $facet->setEmptyBehavior(['behavior' => 'none']);
+      $facet->setThirdPartySetting('facets_pretty_paths', 'coder', 'taxonomy_term_coder');
```
vishalkhode1, line 205: @panshulK We need to add this plugin only if module `facets_pretty_paths` is enabled. Please add that condition.

modules/acquia_cms_search/acquia_cms_search.module
```diff
@@ -161,3 +162,80 @@ function acquia_cms_search_search_api_server_update(EntityInterface $entity) {
...
+  $module_handler->moduleExists('acquia_cms_place') ||
+  $module_handler->moduleExists('acquia_cms_article') ||
+  $module_handler->moduleExists('acquia_cms_event') ||
+  $module_handler->moduleExists('acquia_cms_page')) {
+    // Load the category facet if it exists.
+    $category_facet = Facet::load('search_category');
+    // Create a new facet if the category facet doesn't exists.
+    if (empty($category_facet)) {
+      // Programmatically creating the facet.
+      $facet = Facet::create([
```
vishalkhode1, line 182: @panshulK Let's not add create code here and create a separate class (so we can write PHPUnit tests) and here just add create facet code. (Probably create a class in src/Facade folder and could be name as FacetFacade).

modules/acquia_cms_search/acquia_cms_search.module
```diff
@@ -161,3 +162,80 @@ function acquia_cms_search_search_api_server_update(EntityInterface $entity) {
...
+/**
+ * Implements hook_modules_installed().
+ */
+function acquia_cms_search_modules_installed(array $modules) {
+  $module_handler = Drupal::moduleHandler();
+  // Category facet is created if content type modules are installed.
+  if ($module_handler->moduleExists('acquia_cms_person') ||
+  $module_handler->moduleExists('acquia_cms_place') ||
+  $module_handler->moduleExists('acquia_cms_article') ||
+  $module_handler->moduleExists('acquia_cms_event') ||
```
vishalkhode1, line 175: @panshulK I think we need to add PHPUnit tests as well for all this to make sure when this modules is/are enabled, facet gets created automatically successfully.

vishalkhode1 (changes requested):

modules/acquia_cms_search/acquia_cms_search.module
panshulK: Done

modules/acquia_cms_search/acquia_cms_search.module
panshulK: Done

modules/acquia_cms_search/acquia_cms_search.module
panshulK: Done

modules/acquia_cms_search/tests/src/ExistingSite/AcquiaFacetCreationTest.php
```diff
@@ -0,0 +1,128 @@
...
+    parent::setUp();
+
+    // Setup the Facets and Create Content to test search page functionality.
+    /** @var \Drupal\taxonomy\VocabularyInterface $vocabulary */
+    $vocabulary = Vocabulary::load('categories');
+    $sub_type = $this->createTerm($vocabulary, ['name' => 'Test Category']);
+    $modules = $this->container->get('module_handler');
+
+    // Create facet & content only if the content type module exists.
+    if ($modules->moduleExists('acquia_cms_person')) {
```
panshulK, line 39: There is a better solution than just checking the `acquia_cms_person` module. Doing this, restricts the test to just run only if person content type is installed. 
Instead, we could use a check of the present content types in the system and create content for them all. This will allow the test to run freely with any content type present in the system. Something like this - 

`    $nodeTypes = \Drupal::entityTypeManager()->getStorage('node_type')->loadMultiple();

    if (!empty($nodeTypes)) {
      foreach($nodeTypes as $machine_name => $type) {
        $node = $this->createNode([
          'type' => $machine_name,
          'field_categories' => $sub_type->id(),
          'moderation_state' => 'published',
        ]);
      }
      $facet = Facet::load('test_category_facet');
      if (empty($facet)) {
        // Use the Facade Class to create the facet.
        \Drupal::classResolver(FacetFacade::class)->addFacet('test_category_facet', 'Test Category Facet',
        'category', 'field_categories', 'taxonomy_term_coder', 'search_api:views_page__search__search');
      }
      // Place the facet block in content region to test facets.
      $block = $this->placeBlock('facet_block:test_category_facet', [
        'region' => 'content',
        'id' => 'test_category_facet',
      ]);
      $this->markEntityForCleanup($block);
    }`
    
    Please suggest if this should be done or are we good with just checking the Person content type module.

modules/acquia_cms_search/acquia_cms_search.module
panshulK: Done

modules/acquia_cms_search/acquia_cms_search.module
panshulK: Done

vishalkhode1:
@panshulK I've made following changes in PR:

- Remove ExistingSite PHPUnit tests and added FunctionalJavascript & Kernel tests.
  - Let's not use ExistingSite at all, unless that's the only option or we are writing PHPUnit tests related to Site Studio. Because that might not provided end to end testing modules in isolation.
- Added code to create facets when acquia_cms_search module is enabled after any of acquia cms search dependent modules are installed. (ex: acquia_cms_event, acquia_cms_place etc.)
- Refactored code for FacetFacade class.

cc: @chandan-singh7929

chandan-singh7929 (approved):
LGM

vishalkhode1 (approved):

PR #1017, opened 2021-12-28
ACMS-644: gdpr module updated to 3.0.0-alpha5 release.

vishalkhode1:
**Motivation**
gdpr module updated to 3.0.0-alpha5 release.

**Proposed changes**
Update composer dependencies to gdpr alpha5 version.

**Alternatives considered**
N/A

**Testing steps**
Install and enable acquia_cms_privacy module

**Merge requirements**
- [ ] _Minor change_

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,6 @@
         "drupal/address": "^1",
         "drupal/admin_toolbar": "^3.0",
         "drupal/autologout": "^1",
-        "drupal/checklistapi": "^2.0",
         "drupal/collapsiblock": "^3",
         "drupal/component": "1.0-rc3",
         "drupal/config_ignore": "2.3.0",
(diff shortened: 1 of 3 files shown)


chandan-singh7929 (commented):
@vishalkhode1 - changes looks good to me, one suggestion, If we are patching `drupal/checklistapi` then we should pin it

vishalkhode1:
> @vishalkhode1 - changes looks good to me, one suggestion, If we are patching `drupal/checklistapi` then we should pin it

@chandan-singh7929 Removed the checklist module dependency as we've pinned gdpr module and it's including checlistapi module i.e 2.0.0. So, I think we don't need that module in our composer.json or do we ?

chandan-singh7929:
@vishalkhode1 - Thanks yes that make sense.

Issue #972, opened 2021-10-22
Custom profile creation readme mentioning wrong composer vendor for acms components


joshirohit100:
In https://github.com/acquia/acquia_cms/tree/develop/CUSTOM_PROFILE, composer package has acquia has vendor, for example - `acquia/acquia_cms_article` but this should be `drupal/acquia_cms_article` as this is failing on composer require.

vishalkhode1:
Closing this as we no longer maintaining the acquia_cms distribution.

PR #1028, opened 2022-01-11
Fix composer example in README

jhedstrom:
As noted in #972 the examples using `acquia/acquia_cms_*` should be `drupal/acquia_cms_*`, and the same with `drupal/acquia_claro`.


vishalkhode1:
Closing as we no longer support acquia_cms distribution.

PR #1030, opened 2022-01-13
ACMS-1019: Updated the required configurations.

panshulK:
**Motivation**
* Update configuration files.
Fixes #NNN

**Proposed changes**
* Updated the configuration files as per the ticket requirement.

**Alternatives considered**
* None

**Testing steps**
* Test the site by installing and verifying that everything works as expected.
* Test the required code changes.

**Merge requirements**


vishalkhode1:
@chandan-singh7929 Can you also review this ??

modules/acquia_cms_video/config/dx8/packages.yml
```diff
@@ -1,3 +1,4 @@
 - config/dx8/cohesion_content_templates.media_video_embedded.yml_
 - config/dx8/cohesion_content_templates.media_video_full.yml_
 - config/dx8/cohesion_content_templates.media_video_video_component.yml_
+- config/dx8/cohesion_content_templates.media_image_embedded.yml_
```
chandan-singh7929, line 4: @panshulK - If we are including this shouldn’t we also include for existing site install? using may be update hook.
cc @vishalkhode1

chandan-singh7929:
@panshulK - Running site install I can see following error, we should also remove entry from `packages.yml` file as well
```
 [error]  Message: Cannot access 
> sites/all/modules/acquia_cms_site_studio/config/dx8/cohesion_custom_style.e95bd02c.yml_
>  
>  [error]  Message: Cannot access 
> sites/all/modules/acquia_cms_site_studio/config/dx8/cohesion_custom_style.2b79314d.yml_
```
cc: @vishalkhode1

modules/acquia_cms_video/config/dx8/packages.yml
panshulK: @chandan-singh7929 as discussed the update hook is not required because the file added in acquia_cms_video module is duplicate and has to be removed.

PR #1057, opened 2022-02-18
updating to add support for drush/drush 11.

mikemadison13:
**Motivation**
- Drush 10.x is EOL and we need to support 11.x

**Proposed changes**
- Add ^11 as an allowed dependency for drush/drush


mikemadison13:
rerolled in https://github.com/acquia/acquia_cms/pull/1058

PR #1042, opened 2022-01-31
ACMS-1049: Update to Support Drush 11

rajeshreeputra:
**Motivation**
Fixes #[1049](https://backlog.acquia.com/browse/ACMS-1049)

**Proposed changes**
Update Drush to 11

**Alternatives considered**
NA

**Testing steps**
- Run `composer install`
- Run `drush status` OR `drush -v`
- Check drush version should be updated to 11

**Merge requirements**
- [ ] _Major change_, _Minor change_, _Bug_, _Enhancement_, and/or _Chore_ label applied
- [ ] Manual testing by a reviewer

PR #1058, opened 2022-02-22
ACMS-1049: updating to allow drush 11.

mikemadison13:
Motivation

Drush 10.x is EOL and we need to support 11.x
Proposed changes

Add ^11 as an allowed dependency for drush/drush


vishalkhode1:
Thanks @mikemadison13 for working on this. Closing this PR (as CI tests were failing) as this is now merged with #1042.

PR #1068, opened 2022-03-04
Move non-runtime dependencies to dev

balsama:
**Motivation**
1. Removes the _requirement_ to install non-runtime dependencies from production artifacts (still allows people to install them from their own composer.json file... and gives them much more freedome - see point 2👇)
2. Minimizes potential conflicts (e.g., allows _any version_ of the moved packages to be installed into a project that relies on this package.

**Proposed changes**
Users that relied on this project to install `drush/drush` or `phpspec/prophecy-phpunit` will need to explicitly require those packages on their own.

**Testing steps**
Try to install drush 9 into dev environment. Without this PR, it will fail.


vishalkhode1:
Closing as it seems to be no longer required.

PR #1111, opened 2022-04-08
ACMS-1140: Updated the tour module as per the latest discussion.

panshulK:
**Motivation**
* The tour module isolation test shows /admin/tour/dashboard page as blank page if no helper modules are present.
* Also, the logo is being fetched from acquia_cms profile which is incorrect as the module should work in isolation too.
Fixes #NNN

**Proposed changes**
* Updated code to show a message to user in case no helper modules are present.
* Updated code to fetch the site logo instead of Acquia CMS logo for isolated environments.

**Alternatives considered**
* None

**Testing steps**
* Install Drupal site and create a custom profile.
* install the acquia_cms_tour module on the site and visit the page - /admin/tour/dashboard.
* The dashboard page should be accessible without any errors and should show a message for user to enable helper modules if not enabled.
* If the helper modules are enabled the message to enable will not be shown and the forms would be displayed normally.

For more details, check the recording here - https://www.loom.com/share/0f838724615c4c08a11378eb928030d8

**Merge requirements**


modules/acquia_cms_tour/acquia_cms_tour.module
```diff
@@ -53,3 +53,24 @@ function acquia_cms_tour_modules_uninstalled(array $modules) {
...
+  $showText = 0;
+  $modules = [
+    'google_analytics',
+    'geocoder',
+    'captcha',
+    'cohesion',
+    'google_tag',
+    'acquia_telemetry',
+  ];
+  foreach ($modules as $module) {
```
vishalkhode1, line 70: @panshulK We can't hardcode this module as there could be other plugins which might provide plugin and we should show them. We should get the list from the plugins. See from here: https://github.com/acquia/acquia_cms/blob/develop/modules/acquia_cms_tour/src/Controller/DashboardController.php#L162

modules/acquia_cms_tour/src/Controller/DashboardController.php
```diff
@@ -147,10 +140,19 @@ public function content() {
...
-       are left disabled on install. A checklist is provided to help you keep track of the tasks
-       needed to complete configuration."),
+      The configuration dashboard/wizard setup will help you setup the pre-requisites.
+      Please note, not all modules in ACMS are required by default, and some optional modules
+      are left disabled on install. A checklist is provided to help you keep track of the tasks
+      needed to complete configuration."),
+    ];
+    $form['empty_message'] = [
+      '#type' => 'markup',
+      '#markup' => $this->t("ACMS organizes its features into individual components called modules.
```
vishalkhode1, line 150: @panshulK Can we italicized it ? Also, we should re-write below. Because there could be other plugins as well which depends on other modules.
> Please enable any of the following modules to be able to access the forms
      (ex. google_analytics, google_tag, gecoder, captcha, acquia_telemetry, cohesion)."),

modules/acquia_cms_tour/src/Form/WelcomeModalForm.php
```diff
@@ -66,7 +66,7 @@ public static function create(ContainerInterface $container) {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state, $options = NULL) {
-    $acms_logo = $this->profileExtensionList->getPath('acquia_cms') . '/acquia_cms.png';
+    $logo = file_url_transform_relative(file_create_url(theme_get_setting('logo.url')));
```
vishalkhode1, line 69: @panshulK Both of this functions are deprecated: https://www.drupal.org/node/2940031
We should find alternative and use it. 
**PS:** Make sure the functions are available in Drupal 9.3 & 9.2.

vishalkhode1 (changes requested):

modules/acquia_cms_tour/acquia_cms_tour.module
panshulK: Done!

modules/acquia_cms_tour/src/Controller/DashboardController.php
panshulK: Done!

modules/acquia_cms_tour/src/Form/WelcomeModalForm.php
panshulK: Done!

chandan-singh7929 (approved):
This looks good to me.

PR #1116, opened 2022-04-14
ACMS-1073: Add settings to enable Headless robust api capabilities

modularlab:
**Motivation**
Need to provide a user a starting point for partially / fully headless sites.  

**Proposed changes**
Adds a new service to the ACMS Headless module that when enabled via the Robust API form, completes the following:
- Creates a Headless user Role
- Creates a new Header user
- Creates a consumer entity (with a secret key provided to the user)
- Creates an OAUTH public and key key pair.
- Creates a Next.js site entity with Preview Secret Key
- Gathers a list of available node entities and creates a list of next.js site entity types.
- Also updates the headless.settings config with the base headless User & Consumer UUIDs.

**Alternatives considered**
N / A

**Testing steps**
- Pull branch down locally.
- Do a fresh install of an ACMS site.
- Enable the Acquia CMS Headless module.
- Navigate to the tour page
- enable the robust api option and save
- copy down the secret provided on screen.
- verify that there's a new Headless User and Role, Headless site 1 consumer (default) and Next Js site, as well as Next js entity types matching the available content types.
- Verify the consumer secret is stored, hashed in the database (`drush sql-query "select label, secret from consumer_field_data;"`
- After enabling ACMS Headless and the Robust API feature, see the readme for additional steps for setting up a next js site.

**Merge requirements**


vishalkhode1:
@modularlab Can you please resolve the conflicts.

modularlab:
Merge conflicts addressed.

modularlab:
Merge conflicts addressed.

oauth/.gitignore
```diff
@@ -0,0 +1,2 @@
+# @todo Is this the proper place for oauth keys?
```
vishalkhode1, line 1: @modularlab No, I think this shouldn't be the part of the codebase. This should be out of the codebase ?
cc: @msherron / @fiasco

vishalkhode1:
@modularlab Sorry, can you please rebase your branch once again today (as there was some outage yesterday and we've made some changes), so I can review/merge.
Thannks.

oauth/.gitignore
modularlab: The keys are ignored, and do need to live in the codebase (outside of docroot). but the oauth config does need to be pointed somewhere to reference the keys.

modularlab:
Pulled in the latest and fixed merge conflicts.

oauth/.gitignore
vishalkhode1: @modularlab 

1. I think, since this would be out of drupal `docroot`, then this should be ok. As this path is not web-accessible. Right ? cc: @msherron Opinion ?

2. Also, I think we should create another sub-folder in this(with path or name of current site). Otherwise, if multi-site is setup, then this keys would get overwritten by them. As all multi-site would share the common folder. For this, I think we can use service `'site.path'` as this gives the path for the current site.

3. When I'm installing this module on vanilla drupal (i.e with either standard/minimal profile), then it's giving me following error: 
 ```
The database connection is not serializable. This probably means you are serializing an object that 
has an indirect reference to the database connection. Adjust your code so that is not necessary. 
Alternatively, look at DependencySerializationTrait as a temporary solution. in 
/Users/vishal.khode/Sites/headless/docroot/core/lib/Drupal/Core/Database/Connection.php on line 1993 #0 [internal function]: Drupal\Core\Database\Connection->__sleep()

``` 
When I debugged I found that I don't have `../oauth` directory (in my vanilla drupal codebase). So, I manually created this directory and error is gone.
  3.1 Can we write the code to check & create directory for oauth, in `hook_install()` so that I don't need to manually create directory ?
  3.2 Can we also log the error message in logger, so I can see the error/warnings from DB log ?(https://github.com/acquia/acquia_cms/pull/1116/files#diff-7cc0030c40f70980b5a527c1dea6f2e932c734be1317eb1489bb851a7041571fR197) 

4. When enabling headless mode, I'm seeing the following warning in my DB log. Can we fix this ?
```
Notice: Undefined index: parent in acquia_cms_headless_ui_menu_links_discovered_alter() (line 352 of /Applications/MAMP/htdocs/drupal/docroot/modules/contrib/acquia_cms_headless/modules/acquia_cms_headless_ui/acquia_cms_headless_ui.module)
```

modularlab:
@vishalkhode1,

I made the following updates:

- Fixed error messages related to acquia_cms_headless_ui_menu_links_discovered_alter() updates.
- Added a cache clear to the install/unstall functions for headless ui to make sure that when enabled / disabled the menu system etc. is set as expected.
- Updated how oauth keys are generated.  The directory includes the site.path as part of the file structure.  We are also checking to see if the directory already exists and if not, we are creating it on the users behalf.
- The database serialization issue was actually coming from when we are creating the next js entity types.  We were not checking to see if they already exist.  We can't use duplicate next js site entities.  Now the logic checks to see if the next entity type already exists and if not, it creates it.

PR #1119, opened 2022-04-19
ACMS-1152: Making google analytics as last item in the wizard gives error

rajeshreeputra:
**Motivation**
Fixes #[ACMS-1152](https://backlog.acquia.com/browse/ACMS-1152)

**Proposed changes**
Fix application down error when google analytics is the last item in the tour wizard.

**Alternatives considered**
NA

**Testing steps**

- Install Vanilla Drupal application with:
    - google_analytics
    - acquia_cms_tour
- enable both modules listed above
- Head towards /admin/tour/dashboard
- Click on Wizard set-up
- You should see pop-up with Google Analytics config form
- Click on Skip this step
- You should see the status message(The configuration options have been saved.) and application should not be down 

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_tour/src/Form/InstallationWizardForm.php
+++ b/modules/acquia_cms_tour/src/Form/InstallationWizardForm.php
@@ -271,15 +271,15 @@ class InstallationWizardForm extends FormBase {
    * Skip the current state and mark it as completed.
    */
   public function skipStepSubmit(array &$form, FormStateInterface $form_state) {
-    // Call ignoreConfig of corresponding form.
-    $formController = $this->getCurrentFormController()['class'];
-    $this->classResolver->getInstanceFromDefinition($formController)->ignoreConfig($form, $form_state);
 
     // Call default submitForm in case of last step.
     if ($this->isCurrentStepLast()) {
       $this->submitForm($form, $form_state);
     }
     else {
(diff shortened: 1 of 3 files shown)


modules/acquia_cms_tour/src/Form/InstallationWizardForm.php
```diff
@@ -366,7 +366,9 @@ protected function getFieldValueFromStorage(string $field_name, $empty_value = N
    */
   public function submitForm(array &$form, FormStateInterface $form_state) {
     $formController = $this->getCurrentFormController()['class'];
-    $this->classResolver->getInstanceFromDefinition($formController)->submitForm($form, $form_state);
+    if ($formController != "Drupal\acquia_cms_tour\Plugin\AcquiaCmsTour\GoogleAnalyticsForm") {
```
chandan-singh7929, line 369: @rajeshreeputra - We should not check for specific module here, It could any module so the logic should regardless of any modules config.

rajeshreeputra: Updated as per discussion

chandan-singh7929 (approved):
LGTM.

vishalkhode1 (approved):

PR #1201, opened 2022-08-09
ACMS-1312: Added the patch for the next drupal module changes.

panshulK:
@vishalkhode1 as per the discussion, here's the content save test recording - https://www.loom.com/share/d178c87b53704cb4820579e32016fdf1

**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

**Merge requirements**

I can see the patch being applied correctly.
Screenshot 2022-08-05 at 4 51 52 PM

### DIFF
--- a/modules/acquia_cms_headless/acquia_cms_headless.module
+++ b/modules/acquia_cms_headless/acquia_cms_headless.module
@@ -2,5 +2,24 @@
 
 /**
  * @file
- * Contains hook implementations for the acquia_cms_tour module.
+ * Contains hook implementations for the acquia_cms_headless module.
  */
+
+use Drupal\Core\Url;
+use Drupal\node\NodeInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+
+/**
+ * Implements hook_node_presave().
+ */
+function acquia_cms_headless_node_update(NodeInterface $node) {
+  $url = Url::fromRoute('entity.node.headless_preview', ['node' => $node->id()]);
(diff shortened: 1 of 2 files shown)


vishalkhode1 (approved):

modules/acquia_cms_headless/acquia_cms_headless.module
```diff
@@ -2,5 +2,24 @@
...
+use Symfony\Component\HttpFoundation\RedirectResponse;
+
+/**
+ * Implements hook_node_presave().
+ */
+function acquia_cms_headless_node_update(NodeInterface $node) {
+  $url = Url::fromRoute('entity.node.headless_preview', ['node' => $node->id()]);
+  $headlessConfigurations = \Drupal::config('acquia_cms_headless.settings');
+  // Redirect only when headless mode & next.js starterkit are enabled.
+  if ($headlessConfigurations->get('headless_mode') && $headlessConfigurations->get('starterkit_nextjs')) {
```
fiasco, line 19: @vishalkhode1 is this headless_mode logic used elsewhere? My understanding was that it was essentially a toggle for the acquia_cms_headless_ui module to be enabled. So essentially any headless_mode functionality was placed in the acquia_cms_headless_ui module instead of this module and its functionality was controlled by enabling/disabling the module rather than adding conditional statements like this.

I also don't think this preview behaviour is conditional to the starterkit being enabled. It is nextjs specific but not specific to the use of the starter kit.

PR #1203, opened 2022-08-10
ACMS-1312: Revamped the redirection code for site preview.

panshulK:
**Motivation**
* To remove the re-direction code from acquia_cms_headless module and move to acquia_cms_headless_ui module.
* To re-write the redirection logic to work only when destination params are not set.
* Loom recording - https://www.loom.com/share/8fc6be4203a6428c97c4f98d1b97bb39
Fixes #NNN

**Proposed changes**
* To remove the re-direction code from acquia_cms_headless module and move to acquia_cms_headless_ui module.
* To re-write the redirection logic to work only when destination params are not set.

**Alternatives considered**
* None

**Testing steps**
* Enable headless mode and head towards - /admin/content page.
* Edit any node and notice that node edit URL has the destination query params ex. `?destination=/acquia_cms/admin/content`
* Make changes and notice that the user is redirected to destination url from query params.
* Edit the content again, remove the query params completely and then save.
* Notice that user is redirected to `/node/<id>/site-preview` page.

Loom Recording - https://www.loom.com/share/8fc6be4203a6428c97c4f98d1b97bb39

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_headless/modules/acquia_cms_headless_ui/acquia_cms_headless_ui.module
+++ b/modules/acquia_cms_headless/modules/acquia_cms_headless_ui/acquia_cms_headless_ui.module
@@ -12,9 +12,11 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Render\Element;
 use Drupal\Core\Template\Attribute;
 use Drupal\Core\Url;
+use Drupal\node\NodeInterface;
 use Drupal\node\NodeTypeInterface;
 use Drupal\user\UserInterface;
 use Drupal\views\ViewEntityInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 
 /**
  * Implements hook_help().
@@ -26,6 +28,20 @@ function acquia_cms_headless_ui_help($route_name) {
 
     default:
       return NULL;
+  }
+}
+
+/**
+ * Implements hook_node_update().
+ */
+function acquia_cms_headless_ui_node_update(NodeInterface $node) {
+  $url = Url::fromRoute('entity.node.headless_preview', ['node' => $node->id()]);
+  // Check for destination query param.
+  $current_request = \Drupal::service('request_stack')->getCurrentRequest();
+  // If not set, then redirect to site-preview.
+  if (!$current_request->query->get('destination')) {
+    $response = new RedirectResponse($url->toString());
+    $response->send();
   }
 }
 
@@ -302,7 +318,7 @@ function acquia_cms_headless_ui_menu_links_discovered_alter(array &$links) {
       'admin.content_models',
       'acquia_cms_tour.tour',
       'system.admin_content',
-      'admin_toolbar_tools.help'
+      'admin_toolbar_tools.help',
     ];
     if (in_array($name, $bypass)) {
       continue;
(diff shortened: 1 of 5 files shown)


modules/acquia_cms_headless/modules/acquia_cms_headless_ui/acquia_cms_headless_ui.module
```diff
@@ -29,6 +31,20 @@ function acquia_cms_headless_ui_help($route_name) {
...
+ * Implements hook_node_update().
+ */
+function acquia_cms_headless_ui_node_update(NodeInterface $node) {
+  $url = Url::fromRoute('entity.node.headless_preview', ['node' => $node->id()]);
+  // Check for destination query param.
+  $current_request = \Drupal::service('request_stack')->getCurrentRequest();
+  // If not set, then redirect to site-preview.
+  if (!$current_request->query->get('destination')) {
+    $response = new RedirectResponse($url->toString());
+    $response->send();
```
fiasco, line 44: How is this going to respond in (a) bulk update scenarios and (b) over CLI?

fiasco: This feels like an action we should take on the form rather than on hook_node_update.
This module also has a number or route overrides that maybe worth reviewing. Perhaps the better approach is to set the /node/<nid>/latest route to use entity.node.headless_preview controller?

modules/acquia_cms_headless/modules/acquia_cms_headless_ui/acquia_cms_headless_ui.module
```diff
@@ -96,6 +97,21 @@ function acquia_cms_headless_ui_form_node_form_alter(array &$form, FormStateInte
     ];
   }
   $form['#process'][] = 'acquia_cms_headless_ui_process_node_type_workflow_options';
+  $form['actions']['submit']['#submit'][] = 'acquia_cms_headless_ui_redirect_user_to_site_preview';
```
vishalkhode1, line 100: @panshulK I think the redirection of form is handling from function added above i.e 
`Redirect::entityForm($form, $form_state);`
and looks like handled from [here](https://github.com/acquia/acquia_cms/blob/develop/modules/acquia_cms_headless/modules/acquia_cms_headless_ui/src/Redirect.php#L68).
Please see.

panshulK: Hey @vishalkhode1  I've updated the PR with changes discussed please review.

vishalkhode1 (changes requested):
@panshulK How does the Route Subscriber is getting called ? Are we missing `.serices.yml` file or we've added in some other PR ?

panshulK:
> @panshulK How does the Route Subscriber is getting called ? Are we missing `.serices.yml` file or we've added in some other PR ?
@vishalkhode1  surprisingly i could see the corresponding service.yml file changes when i switched to the file and they were not even showing as "Changes not staged for commit."
I removed the existing service.yml file on local, created new with same contents and then I could see my changes being reflected as "Ready to be staged for commit".

I have updated the PR with the corresponding service file. Please review and let me know if any other changes are required.

Thanks

vishalkhode1 (approved):

PR #1219, opened 2022-08-23
ACMS-1363: Acquia Solr Search should work out of the box when enabled

rajeshreeputra:
**Motivation**
Fixes #ACMS-1363

**Proposed changes**
Acquia Solr Search should work out of the box when enabled, and database server should be disabled.

**Alternatives considered**
NA

**Testing steps**
- install site with this code
- login into application as admin
- head towards /admin/tour/dashboard
- click on Acquia Search accordion 
- provide valid input to all fields
- click on save
- head towards /admin/config/search/search-api
- see Acquia Search Solr Index, Database Search Server should be disabled
- head towards /admin/config/search/search-api/index/content
- check index status

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_search/acquia_cms_search.module
+++ b/modules/acquia_cms_search/acquia_cms_search.module
@@ -159,7 +159,7 @@ function acquia_cms_search_search_api_server_update(EntityInterface $entity) {
 /**
  * Implements hook_form_FORM_ID_alter().
  */
-function acquia_cms_search_form_acquia_cms_solr_search_form_alter(array &$form) {
+function acquia_cms_search_form_acquia_cms_search_form_alter(array &$form) {
   $form['#submit'][] = AcquiaSearchFacade::class . '::submitSettingsForm';
 }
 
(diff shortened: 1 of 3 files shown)


chandan-singh7929 (approved):
@rajeshreeputra - Just a suggestion, I think only updating hook is enough here, no need to disable DB server because it has no index. since index is being attached to solr.
Also, can we make sure, if no connector key is provided it should not update the setting with blank value?
Screenshot 2022-08-24 at 8 17 53 PM

cc: @vishalkhode1

vishalkhode1:
> @rajeshreeputra - Just a suggestion, I think only updating hook is enough here, no need to disable DB server because it has no index. since index is being attached to solr. Also, can we make sure, if no connector key is provided it should not update the setting with blank value? Screenshot 2022-08-24 at 8 17 53 PM
> 
> cc: @vishalkhode1

Correct, I think, it shouldn't switch the server in case of no keys are provided ?

chandan-singh7929 (approved):
LGTM

PR #1237, opened 2022-08-31
ACMS-1384: Fix the failing PHPunit tests.

vishalkhode1:
**Motivation**
Fixes failing CI tests.

**Proposed changes**
There were 2 issues mainly:

1. When any of acquia_cms content model gets installed, it creates the roles that are required i.e content_author, content_editor, content_administrator, but it only adds the permissions related to that content model. Ex: If we enable acquia_cms article, then roles gets the permissions: create article content, edit article content, delete article content , but they do not get the other required default permissions i.e use text format, clone entity etc. and due to this those 3 Embed tests were failing.
2. There were certain optional config which depends on this roles. But now we are dynamically creating role, so this optional configuration were not getting imported and thus workbench_email config not getting imported and causing to fail the tests.

**Alternatives considered**
I did thought of alternative to fix point **2.)** from above proposed changes. Currently, I've added fix that: Whenever we are creating role, after any of content model module gets installed, we are running Drupal core service to import the optional configurations which are dependent on this roles and thus the optional configuration gets imported.
The alternative I had in my mind is:
Instead of using `hook_module_installed()`, maybe we can use `hook_preinstall()` and create the roles. Because after enabling any module, Drupal also see if there are any optional configurations which it can import, so thus way I don't need to add the code (which I added now) to import optional configurations.

**Testing steps**
- PHPUnit tests should pass
- Manual Testing.

**Merge requirements**
- [ ] _Major change_


chandan-singh7929:
@vishalkhode1 - There are some other failure that I think we should fix:
https://app.travis-ci.com/github/acquia/acquia_cms/jobs/581477388#L12823

modules/acquia_cms_common/acquia_cms_common.module
```diff
@@ -340,11 +340,17 @@ function acquia_cms_common_modules_installed($modules) {
...
       'acquia_cms_event',
       'acquia_cms_page',
       'acquia_cms_person',
       'acquia_cms_place',
+      // Ensure role exists when any of media module is installed.
+      'acquia_cms_document',
+      'acquia_cms_audio',
+      'acquia_cms_video',
+      'acquia_cms_image',
     ])) {
```
chandan-singh7929, line 354: @vishalkhode1 - Please move this media module dependency for creating content role, rather we should use permission facade to append required permissions to the role if exists.

modules/acquia_cms_common/acquia_cms_common.module
```diff
@@ -355,6 +361,22 @@ function acquia_cms_common_modules_installed($modules) {
...
+    // So we need to make sure that we also import the other optional
+    // configurations which is dependent on any of these roles.
+    /** @var \Drupal\Core\Config\ConfigInstallerInterface $config_installer */
+    $config_installer = \Drupal::service('config.installer');
+    $restrict_by_dependency = [
+      'config' => 'user.role.content_administrator',
+      'config' => 'user.role.content_author',
+      'config' => 'user.role.content_editor',
+    ];
+    $config_installer->installOptionalConfig(NULL, $restrict_by_dependency);
```
chandan-singh7929, line 379: @vishalkhode1 - If possible can use use module preinstall hook for the same?

chandan-singh7929 (changes requested):
Requested some changes please see.

PR #1241, opened 2022-09-05
ACMS-1378: Added code to enable syslog module on IDE env.

panshulK:
**Motivation**
* Syslog Module should be enabled based on different environments.
Fixes #NNN

**Proposed changes**
* Managed the conditions for enabling the syslog module on different environments.

**Alternatives considered**
* None

**Testing steps**
* Install site using command - `vendor/bin/drush site:install minimal --account-pass admin -vvv`
* Once the site is installed, run the command - `vendor/bin/drush acms:toggle:modules` and verify that syslog module is enabled / disabled according to the criteria mentioned in the [ticket](https://backlog.acquia.com/browse/ACMS-1378).

Test 2
* Install site using `vendor/bin/acms acms:install` command.
* Visit site & notice that syslog module enablement/disablement is handled on it's own according to the environment installed on.

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_common/acquia_cms_common.install
+++ b/modules/acquia_cms_common/acquia_cms_common.install
@@ -75,6 +75,10 @@ function acquia_cms_common_install() {
     $config = \Drupal::service('purge.purgers');
     $config->setPluginsEnabled(['cee22bc3fe' => 'acquia_purge']);
   }
+
+  // Install the syslog module.
+  $module_installer = \Drupal::service('module_installer');
+  $module_installer->install(['syslog']);
 }
 
 /**
--- a/modules/acquia_cms_common/src/Services/ToggleModulesService.php
+++ b/modules/acquia_cms_common/src/Services/ToggleModulesService.php
@@ -13,17 +13,16 @@ class ToggleModulesService {
    * Toggle module based on environment.
    */
   public function toggleModules() {
-    $is_dev = Environment::isLocalEnv();
+    $is_dev = Environment::isAhIdeEnv() || Environment::isLocalEnv();
     $is_prod = Environment::isAhProdEnv();
     $module_installer = \Drupal::service('module_installer');
-    $to_install = [];
     $to_uninstall = [];
     if ($is_dev) {
       array_push($to_install, 'dblog', 'jsonapi_extras', 'field_ui', 'views_ui');
-      array_push($to_uninstall, 'syslog', 'autologout');
+      array_push($to_uninstall, 'autologout');
     }
     else {
-      array_push($to_install, 'syslog', 'autologout');
+      array_push($to_install, 'autologout');
     }
     if (!$is_prod) {
       array_push($to_install, 'reroute_email');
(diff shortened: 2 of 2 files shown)


fiasco:
This PR suggests that syslog shouldn't be enabled on dev environments, however, it would be useful on docker based dev environments like Lando, DDEV and docksal. I suggest we revise to ensure syslog is always enabled. I don't think there is a real scenario where it is dangerous to have it on?

vishalkhode1 (approved):

vishalkhode1:
> This PR suggests that syslog shouldn't be enabled on dev environments, however, it would be useful on docker based dev environments like Lando, DDEV and docksal. I suggest we revise to ensure syslog is always enabled. I don't think there is a real scenario where it is dangerous to have it on?

@fiasco I think customers/users mostly use db log in local and syslog in Acquia environment. That's reason we are not enforcing syslog to enable every time. If users want to change from dblog to syslog, they can do in their local and send the updated configuration.

fiasco:
Is there an issue with having both modules on in dev? I appreciate the value of dblog in local dev environments but its not clear to me why it'd be valuable to explicitly turn syslog off in local dev. Couldn't we just have it enabled everywhere by default? It doesn't matter if it doesn't work in local dev (similar to how it doesn't matter if sendmail is not configured in local dev). But if it was, it would remove reduce the amount of environment specific configuration to manage and solve this particular issue.

vishalkhode1:
> Is there an issue with having both modules on in dev?

I would say yes. Because Acquia recommends to use syslog in favor of dblog, specifically on Acquia Cloud environment. See [here](https://docs.acquia.com/cloud-platform/develop/drupal/module-caution/). So, I wouldn't recommend to turn on both modules.


> Couldn't we just have it enabled everywhere by default? It doesn't matter if it doesn't work in local dev (similar to how it doesn't matter if sendmail is not configured in local dev). But if it was, it would remove reduce the amount of environment specific configuration to manage and solve this particular issue.

Yes, we can do that as well. But if enable syslog in local (and not dblog). Then customers/users have to explicitly enable dblog and see what errors/issues they are getting (If they haven't configured syslog). So, they would expect that at-least dblog module should be enabled in local, so they can quickly see what errors/issues they are getting. And then they know the recommendations that Acquia provides, so they would enable syslog, send the updated configurations and run drush cim. Also, if they need, they can use config split and use different modules based on their needs.

modules/acquia_cms_common/acquia_cms_common.info.yml
```diff
@@ -44,3 +44,4 @@ dependencies:
   - drupal:taxonomy
   - drupal:username_enumeration_prevention
   - drupal:workbench_email
+  - drupal:syslog
```
panshulK, line 47: @vishalkhode1 need your feedback here. As per my understanding, the syslog module should be enabled by default. So it should be enabled on site install? If yes, then hence the module is mentioned here. Else the file changes in `ToggleMofuleService.php` file below should help achieve the required functionality.

Thanks

vishalkhode1: @panshulK This looks correct to me though, but for now let's enable only for the new customers who are using that toggle service. Though, I'm unsure. @fiasco can you suggest your opinion on this ?

vishalkhode1 (approved):

PR #1249, opened 2022-09-08
ACMS-1404 : Update Next to latest with new patch

deepakmishra2:
**Motivation**

Fixes ACMS-1404

**Proposed changes**

**Alternatives considered**

**Testing steps**

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_headless/acquia_cms_headless.install
+++ b/modules/acquia_cms_headless/acquia_cms_headless.install
@@ -5,6 +5,9 @@
  * Install, update and uninstall functions for the acquia_cms_headless module.
  */
 
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\user\RoleInterface;
+
 /**
  * Implements hook_install().
  */
@@ -14,6 +17,10 @@ function acquia_cms_headless_install() {
 
   // Programmatically create a new "headless" user.
   $headless_user->createHeadlessUser();
+
+  // Adds the necessary permissions to roles.
+  add_next_site_permissions();
+
 }
 
 /**
@@ -32,4 +39,47 @@ function acquia_cms_headless_requirements($phase) {
   }
 
   return $requirements;
+}
+
+/**
+ * Implements hook_update_N().
+ *
+ * Update all consumers and assign content model roles and add
+ * provide `issue subrequests` permission to authenticated & anonymous role.
+ */
+function acquia_cms_headless_update_8001() {
+  $rolesStorage = \Drupal::entityTypeManager()->getStorage('user_role');
+  $query = $rolesStorage->getQuery();
+  $roleIds = $query
+    ->condition('id', ["authenticated", "anonymous", "headless"], 'NOT IN')
+    ->execute();
+  $config = \Drupal::configFactory()->getEditable('acquia_cms_headless.settings');
+  if ($config->get('consumer_uuid')) {
+    $consumer = \Drupal::service('entity.repository')->loadEntityByUuid("consumer", $config->get('consumer_uuid'));
+    if ($consumer instanceof EntityInterface && $consumer->user_id->entity instanceof EntityInterface) {
+      $user = $consumer->user_id->entity;
+      foreach ($roleIds as $roleId) {
+        $user->addRole($roleId);
+      }
+      $user->save();
+    }
+  }
+  add_next_site_permissions();
+}
+
+/**
+ * Adds the necessary permissions to anonymous & authenticated roles.
+ *
+ * @throws \Drupal\Core\Entity\EntityStorageException
+ */
+function add_next_site_permissions() {
+  $rolesStorage = \Drupal::entityTypeManager()->getStorage('user_role');
+  $roles = $rolesStorage->loadMultiple([
+    RoleInterface::ANONYMOUS_ID,
+    RoleInterface::AUTHENTICATED_ID,
+  ]);
+  foreach ($roles as $role) {
+    $role->grantPermission('issue subrequests');
+    $role->save();
+  }
 }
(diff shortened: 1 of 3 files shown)


rajeshreeputra (approved):
Everything looks good, we observed that preview is loading but not completing even after long time, NextJs site working fine.

lauriii (changes requested):
Please review the proposed steps from ACMS-1404. Those are critical and this issue _must not_ be merged before that because the steps are required to mitigate [SA-CONTRIB-2022-054](https://www.drupal.org/sa-contrib-2022-054). 

I believe we should also increase the version constraint for `drupal/next` to `^1.3.0` which will ensure that all sites are updated.

chandan-singh7929:
Yes, I can confirm that the preview is not working, its keep loading.

lauriii:
> Yes, I can confirm that the preview is not working, its keep loading.

Have you tried following the steps from the Jira issue? That should fix the preview.

vishalkhode1:
> > Yes, I can confirm that the preview is not working, its keep loading.
> 
> Have you tried following the steps from the Jira issue? That should fix the preview.

@lauriii Do you mean point #2, #3 & #4 from [this](https://docs.google.com/document/d/1_acd12Qu64BOrhnA1RywABefthj6mPPwh3mDL4pIaMA/edit) document ?

lauriii:
Yes. On top of that you will have to do steps from "Changes needed in next-acms" locally on your Next.js application

modules/acquia_cms_headless/acquia_cms_headless.install
```diff
@@ -33,3 +39,49 @@ function acquia_cms_headless_requirements($phase) {
...
   return $requirements;
 }
+
+/**
+ * Implements hook_update_N().
+ *
+ * Update all consumers and assign content model roles and add
+ * provide `issue subrequests` permission to authenticated & anonymous role.
+ */
+function acquia_cms_headless_update_8001() {
```
lauriii, line 49: This should probably be a post_update hook. These updates are executed after all hook_update_n which ensures that Drupal is already fully repaired which makes it safe to update the entities.

modules/acquia_cms_headless/acquia_cms_headless.install
```diff
@@ -33,3 +39,49 @@ function acquia_cms_headless_requirements($phase) {
...
+    ->condition('id', ["authenticated", "anonymous", "headless"], 'NOT IN')
+    ->execute();
+
+  $consumerStorage = \Drupal::entityTypeManager()->getStorage('consumer');
+  $query = $consumerStorage->getQuery();
+  $consumerIds = $query
+    ->exists('user_id')
+    ->execute();
+  $consumers = $consumerStorage->loadMultiple($consumerIds);
+  foreach ($consumers as $consumer) {
```
lauriii, line 62: Are we sure we want to do this was all consumer users because sites could have consumers that are used for other purposes? I think we should do this only for the default consumer created by ACMS.

modules/acquia_cms_headless/src/Service/StarterkitNextjsService.php
```diff
@@ -298,7 +298,9 @@ public function createHeadlessUser() {
         $user->set("init", $email);
         $user->set("preferred_langcode", $language);
         $user->activate();
-        $user->addRole('headless');
+        foreach ($this->getAllSiteRoles() as $role) {
```
lauriii, line 301: I think this could benefit from a comment that explains why we are granting all of the roles for this user.

lauriii (changes requested):
I reviewed the code but I haven't been able to install a development environment for `acquia_cms` so I can't test this locally.

modules/acquia_cms_headless/acquia_cms_headless.install
```diff
@@ -33,3 +39,49 @@ function acquia_cms_headless_requirements($phase) {
...
+    $user = $consumer->user_id->entity;
+    foreach ($roleIds as $roleId) {
+      $user->addRole($roleId);
+    }
+    $user->save();
```
joshirohit100, line 67: will this also impact more consumers/users than we expecting? What if a customer site has a consumer for specific purpose and this change will add admin role/scope to that

modules/acquia_cms_headless/acquia_cms_headless.install
vishalkhode1: @lauriii We are not updating config entities, we are updating content entity (i.e Headless user), Also, we are adding new permission to role config entity (permission config to role entity is provide by CORE), so I think hook_update_N() should be fine.

modules/acquia_cms_headless/acquia_cms_headless.install
vishalkhode1: Ok, will update this.

modules/acquia_cms_headless/acquia_cms_headless.install
lauriii: I believe content entities also should be updated in a post_update for the same reasons as config entities. This _could_ be fine but post_update should be safer.

lauriii:
Tested manually for a new installation and was getting the results I was hoping. 👍 Haven't been able to test this on an existing site because the steps for creating a development environment required me to reinstall ACMS. Could someone who has pre-existing development environment confirm that the update path works (potentially after https://github.com/acquia/acquia_cms/pull/1249#discussion_r965857462 has been addressed)?

modules/acquia_cms_headless/acquia_cms_headless.install
lauriii: That depends on which modules and module versions a site is running, and which updates needs to be run. Like I said, this _could_ be fine but if you want to be sure, you should use post_update for this. The hook_update_n is supposed to be used for database schema changes and other low level changes like that, which are critical for the site to function. Updating entities doesn't fall into that category.

vishalkhode1: For now, I'm merging this as this needs to be released ASAP. Will look later in detail and will see if this needs to be updated.

PR #1261, opened 2022-09-15
ACMS-1380: Removed node_revision_delete config from node.type.page

rajeshreeputra:
**Motivation**
Fixes #ACMS-1380

**Proposed changes**
- Remove node_revision_delete config from acquia_cms_page/config/optional/node.type.page.yml
- Add update hook to remove node_revision_delete config from node.type.page

**Alternatives considered**
NA

**Testing steps**
**New site:**
- Install site with this branch
- Install acquia_cms_page, node_revision_delete config shouldn't be there 
- Install acquia_cms_site_studio, node_revision_delete third party config must be available
- Follow 2 and 3 in reverse manner with new install
- Verify site working fine

**Existing site:**
- Upgrade codebase this branch
- Execute `drush updb`
- Now execute `drush cex` so node_revision_delete does not exist in your exported configs
- Now execute `drush cim`
- Verify site working fine

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_site_studio/acquia_cms_site_studio.install
+++ b/modules/acquia_cms_site_studio/acquia_cms_site_studio.install
@@ -84,6 +84,19 @@ function acquia_cms_site_studio_install() {
       ->save();
   }
 
+  // Set default node revision delete configuration for Page content type.
+  $moduleHandler = \Drupal::service('module_handler');
+  if ($moduleHandler->moduleExists('acquia_cms_page')) {
+    $type = NodeType::load('page');
+    $third_party_settings = $type->get('third_party_settings');
+    if ($type && !isset($third_party_settings['node_revision_delete'])) {
+      $type->setThirdPartySetting('node_revision_delete', 'minimum_revisions_to_keep', '30');
+      $type->setThirdPartySetting('node_revision_delete', 'minimum_age_to_delete', '0');
+      $type->setThirdPartySetting('node_revision_delete', 'when_to_delete', '0');
+      $type->save();
+    }
+  }
+
   // Re-write the editor and filter format config from
   // optional directory since we have don some update to it.
   $module_path = \Drupal::moduleHandler()->getModule('acquia_cms_site_studio')->getPath();
--- a/modules/acquia_cms_site_studio/acquia_cms_site_studio.module
+++ b/modules/acquia_cms_site_studio/acquia_cms_site_studio.module
@@ -14,7 +14,6 @@ use Drupal\cohesion_website_settings\Controller\WebsiteSettingsController;
 use Drupal\Core\Ajax\CloseDialogCommand;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\media_library\MediaLibraryState;
-use Drupal\node\Entity\NodeType;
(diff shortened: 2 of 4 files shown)


modules/acquia_cms_site_studio/acquia_cms_site_studio.module
```diff
@@ -213,3 +214,20 @@ function acquia_cms_site_studio_rebuild_styles() {
   // rebuild.
   batch_set(install_acms_site_studio_rebuild());
 }
+
+/**
+ * Implements hook_modules_installed().
+ */
+function acquia_cms_site_studio_modules_installed($modules) {
+  // Set default node revision delete configuration for Place content type.
```
chandan-singh7929, line 222: @rajeshreeputra - Update comment here, it should be page not place

chandan-singh7929 (approved):
Overall looks good to me, there's minor update I have requested.

modules/acquia_cms_site_studio/acquia_cms_site_studio.module
rajeshreeputra: updated

chandan-singh7929 (approved):

modules/acquia_cms_site_studio/acquia_cms_site_studio.module
```diff
@@ -213,3 +214,20 @@ function acquia_cms_site_studio_rebuild_styles() {
   // rebuild.
   batch_set(install_acms_site_studio_rebuild());
 }
+
+/**
+ * Implements hook_modules_installed().
+ */
+function acquia_cms_site_studio_modules_installed($modules) {
```
vishalkhode1, line 221: @rajeshreeputra We shouldn't use this hook as this hooks gets invoked each time a module is installed. For ex: If I install 20 modules one by one, then will be called 20 times, but we don't need to update config 20 times. Please use hook_module_install() and also, add condition check for `node_revision_delete` module only.
cc: @chandan-singh7929

rajeshreeputra: updated

modules/acquia_cms_site_studio/acquia_cms_site_studio.install
```diff
@@ -83,6 +83,18 @@ function acquia_cms_site_studio_install() {
       ->save();
   }
 
+  // Set default node revision delete configuration for Page content type.
+  $moduleHandler = \Drupal::service('module_handler');
+  if ($moduleHandler->moduleExists('acquia_cms_page')) {
```
vishalkhode1, line 88: @rajeshreeputra Can we the same condition check as above i.e `node_revision_delete` ? As it would cause issue, where anyone using old `acquia_cms_page` module which doesn't add dependency for node_revision_delete module.

vishalkhode1 (changes requested):

modules/acquia_cms_site_studio/acquia_cms_site_studio.install
rajeshreeputra: Will add is config empty condition here.

vishalkhode1 (approved):

PR #1287, opened 2022-10-10
ACMS-1480: Added test case for headless installation frontpage checks.

panshulK:
**Motivation**
* Added test cases for headless installation frontpage checks.
Fixes #NNN

**Proposed changes**
* Added test cases for headless installation frontpage checks. The test case checks for the following -
1. Frontpage is login when not authenticated.
2. Frontpage is content admin page when logged in.
3. Frontpage theme is admin theme.

**Alternatives considered**
* None

**Testing steps**
* Switch to branch ACMS-1480 on local codebase.
* Run PHPUnit tests using following command - `./vendor/bin/phpunit -c docroot/core modules/acquia_cms_headless/modules/acquia_cms_headless_ui/tests/src/ExistingSite/AcquiaCMSHeadlessFrontpageTest.php --debug -v`

**Merge requirements**

Screenshot 2022-10-10 at 3 58 57 PM

### DIFF
--- a/modules/acquia_cms_headless/modules/acquia_cms_headless_ui/tests/src/Functional/AcquiaCMSHeadlessFrontpageTest.php
+++ b/modules/acquia_cms_headless/modules/acquia_cms_headless_ui/tests/src/Functional/AcquiaCMSHeadlessFrontpageTest.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Drupal\Tests\acquia_cms_headless_ui\Functional;
+
+use Drupal\Tests\BrowserTestBase;
+
+/**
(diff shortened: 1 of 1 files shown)


chandan-singh7929:
@panshulK - I think we should use Functional test instead of ExistingSite.
cc: @vishalkhode1

vishalkhode1:
> @panshulK - I think we should use Functional test instead of ExistingSite. cc: @vishalkhode1

@panshulK Yes, we shouldn't use ExistingSite tests for any module (only exception is in case of adding tests for Site Studio functionalities).

vishalkhode1:
Closing as w'll be not using ExistingSite test for acquia_cms_headless module.

PR #1307, opened 2022-10-28
ACMS-1508: Acquia telemetry cleanup and update from acquia cms modules.

apathak18:
**Motivation**

Fixes ACMS-1508

**Proposed changes**
Cleaning up the acquia telemetry reference in acquia cms modules.

**Testing steps**
Check the dashboard page via tour
No separate acquia telemetry form should be there.


modules/acquia_cms_search/src/Plugin/AcquiaCmsTour/AcquiaConnectorForm.php
```diff
@@ -38,6 +40,32 @@ protected function getEditableConfigNames() {
...
+   * The site profile.
+   *
+   * @var \Drupal\acquia_connector\SiteProfile\SiteProfile
+   */
+  protected SiteProfile $siteProfile;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
```
vishalkhode1, line 60: I think we can rewrite this whole as:
```suggestion
  public static function create(ContainerInterface $container) {
    $this->siteProfile = $container->get('acquia_connector.site_profile');
    $this->subscription = $container->get('acquia_connector.subscription');
    return parent::create($container);
  }
```

vishalkhode1 (changes requested):

modules/acquia_cms_search/src/Plugin/AcquiaCmsTour/AcquiaConnectorForm.php
apathak18: $this-> reference isn't allowed

rajeshreeputra (approved):
Changes looks good to me, please move telemetry config to common

modules/acquia_cms_search/src/Plugin/AcquiaCmsTour/AcquiaConnectorForm.php
```diff
@@ -60,15 +91,24 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
         '#type' => 'textfield',
         '#title' => $this->t('Name'),
         '#maxlength' => 255,
         '#disabled' => TRUE,
         '#required' => TRUE,
-        '#default_value' => $this->state->get('spi.site_name'),
-        '#prefix' => '<div class= "dashboard-fields-wrapper">' . $module_info['description'],
-        '#suffix' => "</div>",
+        '#default_value' => $this->state->get('spi.site_name') ?? $site_name,
+        '#suffix' => '</div class= "dashboard-fields-wrapper">',
```
rajeshreeputra, line 111: any reason for removal of prefix here?

apathak18: Yes to make UI like this.
#prefix attribute is moved to checkbox field.
Screenshot 2022-11-04 at 6 41 25 PM

modules/acquia_cms_common/acquia_cms_common.module
```diff
@@ -437,6 +437,13 @@ function acquia_cms_common_modules_installed($modules) {
       \Drupal::service('cohesion_sync.package_import_handler')->importPackagesFromPath($packages);
     }
   }
+
+  // Acquia connector check and set telemetry opt-in feature.
+  if (\Drupal::moduleHandler()->moduleExists('acquia_connector') &&
```
vishalkhode1, line 442: @apathak18 I think there's no need to check this on `hook_modules_installed()` instead we should add this in `hook_install()`.

modules/acquia_cms_common/acquia_cms_common.module
```diff
@@ -456,3 +463,18 @@ function acquia_cms_common_preprocess_maintenance_page(array &$variables) {
...
 }
+
+/**
+ * Implements hook_module_implements_alter().
+ */
+function acquia_cms_common_module_implements_alter(array &$implementations, string $hook): void {
+  if (\Drupal::moduleHandler()->moduleExists('acquia_connector') &&
+  ($hook === 'modules_installed' || $hook === 'modules_uninstalled')) {
+    // Check if telemetry not opted.
+    if (!\Drupal::state()->get('acquia_connector.telemetry.opted')) {
```
vishalkhode1, line 474: @apathak18 Let's pass the second argument to `state()->get()` as TRUE which opt in the user, if no value exist.
```suggestion
    if (!\Drupal::state()->get('acquia_connector.telemetry.opted', TRUE)) {
```
Let's do the same in eventSubscriber as well (unless you see any issue).

PR #1331, opened 2022-11-24
ACMS-1542: Support for Site Studio 7.0.x

chandan-singh7929:
**Motivation**

Fixes #ACMS-1542

**Proposed changes**

We are re-exporting site studio packages as per 7.x

**Alternatives considered**

**Testing steps**

- Checkout to ACMS-1542
- Run composer require drupal/core:^9.5
- Site install with site studio key present ./vendor/bin/acms si
- Update text format site studio to use ckeditor5
- Create page with any component using text editor and make sure its working as expected.

**Merge requirements**


composer.json
```diff
@@ -18,6 +18,8 @@
     ],
     "require": {
         "acquia/acquia-cms-starterkit": "^1.0",
+        "acquia/cohesion": "7.0.x-dev",
```
vishalkhode1, line 21: @deepakmishra2 Can we please revert this ? As this shouldn’t be updated as part of this ticket.

.github/workflows/acquia_cms_ci.workflow.yml
```diff
@@ -98,7 +98,7 @@ jobs:
     strategy:
       matrix:
         orca-job:
-          - ISOLATED_TEST_ON_CURRENT_DEV
+          - ISOLATED_TEST_ON_NEXT_MINOR
```
vishalkhode1, line 101: @deepakmishra2 Please revert this (and below as well) as we should test on Drupal core 9.5 seperately.

vishalkhode1 (changes requested):

PR #1345, opened 2022-12-02
ACMS-1555: Fixing null deprecations for site studio 6.9.x

apathak18:
**Motivation**

Fixes ACMS-1555

**Proposed changes**
Update the patch to fix the null deprecations warnings.

**Testing steps**

- Add a page content type
- Add a component from site studio.


rajeshreeputra (approved):
Looks good to me

PR #1352, opened 2022-12-08
ACMS-1567: Removing hard dependency of later runtime processor and move into hook_install.

apathak18:
**Motivation**

Fixes #ACMS-1567

**Proposed changes**
Remove  later runtime processor from info.yml and move into hook_install

### DIFF
--- a/modules/acquia_cms_common/acquia_cms_common.install
+++ b/modules/acquia_cms_common/acquia_cms_common.install
@@ -60,9 +60,10 @@ function acquia_cms_common_install() {
   $config->set('enable_export_filtering', TRUE);
   $config->save(TRUE);
 
-  // Uninstall page_cache module in favour of memcache.
+  $module_installer = \Drupal::service('module_installer');
   if ($module_handler->moduleExists('page_cache')) {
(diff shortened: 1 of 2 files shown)


rajeshreeputra (approved):
Looks good to me.

modules/acquia_cms_common/acquia_cms_common.install
```diff
@@ -78,8 +79,9 @@ function acquia_cms_common_install() {
   }
 
   // Install the syslog module.
-  $module_installer = \Drupal::service('module_installer');
   $module_installer->install(['syslog']);
```
vishalkhode1, line 82: @apathak18 No need to call install method again below, we can update the same function i.e
```suggestion
  $module_installer->install(['syslog', 'purge_processor_lateruntime']);
```

vishalkhode1: Also, I think we should do the same for purge_processor_cron & purge_ui module.
cc: @rajeshreeputra / @chandan-singh7929

apathak18: Done

PR #1371, opened 2022-12-21
ACMS-1603: Move profile switch patch from common module to acquia_cms profile.

rajeshreeputra:
**Motivation**
Fixes #ACMS-1603

**Proposed changes**
Move profile switch patch from common module to acquia_cms profile.

**Alternatives considered**
NA

**Testing steps**
- Upgrade to OR download latest version of acquia_cms profile 
- verify patch is applied on core successfully
- verify profile switch command works fine

**Merge requirements**
- [ ] _Major change_, _Minor change_, _Bug_, _Enhancement_, and/or _Chore_ label applied
- [ ] Manual testing by a reviewer

PR #1372, opened 2022-12-21
ACMS-1603: Removed profile switch core patch.

rajeshreeputra:
**Motivation**
Fixes #ACMS-1603 

**Proposed changes**
Removed profile switch core patch.

**Alternatives considered**
NA

**Testing steps**
- Checkout ACMS-1603-common branch `git checkout ACMS-1603-common`
- Execute `composer install`
- Verify that profile switch patch does not present and getting applied

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_common/composer.json
+++ b/modules/acquia_cms_common/composer.json
@@ -64,7 +64,8 @@
         "drupal/acquia_cms_starter": "<1.3.3",
         "drupal/acquia_cms_toolbar": "<1.4",
         "drupal/acquia_cms_tour": "2.0.* || <1.3.2",
-        "drupal/acquia_cms_video": "<1.4"
+        "drupal/acquia_cms_video": "<1.4",
+        "acquia/acquia_cms": "<1.5.2"
     },
     "config": {
         "allow-plugins": {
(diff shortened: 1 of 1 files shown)


apathak18 (approved):
Changes looks good to me

vishalkhode1 (approved):

modules/acquia_cms_common/composer.json
```diff
@@ -64,7 +64,8 @@
         "drupal/acquia_cms_starter": "<1.3.3",
         "drupal/acquia_cms_toolbar": "<1.4",
         "drupal/acquia_cms_tour": "2.0.* || <1.3.2",
-        "drupal/acquia_cms_video": "<1.4"
+        "drupal/acquia_cms_video": "<1.4",
+        "acquia/acquia_cms": "<1.5.2"
```
vishalkhode1, line 68: @rajeshreeputra I think this is not required. As we are already adding conflict [here](https://github.com/acquia/acquia_cms/pull/1371/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R42).

rajeshreeputra: We have added in both version because we releases acquia_cms_common next version before 1.5.x then there will be no patch available. To make sure core patch existing after upgrade we have added in both places.

PR #1381, opened 2022-12-28
ACMS-1611: Fix headless_ui issue for admin block and access check

deepakmishra2:
**Motivation**

Fixes #ACMS-1611 and ACMS-1610

**Proposed changes**

**Alternatives considered**

**Testing steps**

**Merge requirements**


modules/acquia_cms_headless/src/Plugin/AcquiaCmsHeadless/HeadlessApiKeys.php
```diff
@@ -128,7 +128,7 @@ public function getEntityData() {
     $query->tableSort($header);
     $query->pager(10);
 
-    return $query->execute();
+    return $query->accessCheck(FALSE)->execute();
```
apathak18, line 131: @deepakmishra2 can we update with **accessCheck(TRUE)** ??

apathak18 (changes requested):
accessCheck(TRUE) for all queries execution in our function.
Please do the same for all because I don't think we are allowing anonymous user to access these queries output.

cc: @vishalkhode1

modules/acquia_cms_headless/src/Plugin/AcquiaCmsHeadless/HeadlessApiKeys.php
deepakmishra2: done

apathak18 (approved):
LGTM
Existing test failures fixes is being done in PR-1374

PR #1419, opened 2023-01-30
ACMS-1480: acquia_cms_headless: PHPUnit testcases for Frontpage

deepakmishra2:
Motivation
Fixes https://github.com/acquia/acquia_cms/pull/1287

Proposed changes
Added test cases for headless installation frontpage checks. The test case checks for the following -
Frontpage is login when user is not authenticated.
Frontpage is content admin page when user logged in with admin.
Frontpage theme is admin theme.

### DIFF
--- a/modules/acquia_cms_headless/modules/acquia_cms_headless_ui/tests/src/Functional/HeadlessFrontpageTest.php
+++ b/modules/acquia_cms_headless/modules/acquia_cms_headless_ui/tests/src/Functional/HeadlessFrontpageTest.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Drupal\Tests\acquia_cms_headless_ui\Functional;
+
+use Drupal\Tests\acquia_cms_headless\Functional\HeadlessBrowserTestBase;
+
+/**
+ * Tests for acquia_cms_headless frontpage.
+ *
+ * @group acquia_cms_headless
+ * @group low_risk
+ */
+class HeadlessFrontpageTest extends HeadlessBrowserTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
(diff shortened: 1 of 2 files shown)


apathak18 (approved):
LGTM

modules/acquia_cms_headless/modules/acquia_cms_headless_ui/tests/src/Functional/HeadlessFrontpageTest.php
```diff
@@ -0,0 +1,76 @@
...
+use Acquia\DrupalEnvironmentDetector\AcquiaDrupalEnvironmentDetector;
+use Drupal\Tests\BrowserTestBase;
+
+/**
+ * Tests for acquia_cms_headless frontpage.
+ *
+ * @group acquia_cms_headless
+ * @group low_risk
+ */
+class HeadlessFrontpageTest extends BrowserTestBase {
```
apathak18, line 14: @deepakmishra2 can we extend DashboardBrowserTestBase instead of BrowserTestBase, if yes then it would save the code repeatition.

cc: @rajeshreeputra @vishalkhode1

deepakmishra2: @apathak18 DashboardBrowserTestBase uses DashboardSectionTrait i dont think we need that trait, also i need UI module to test this feature whereas in DashboardBrowserTestBase it uses entire headless module.

cc @vishalkhode1

rajeshreeputra (approved):
please see comments in ticket

rajeshreeputra (approved):
Looks good to me

PR #1513, opened 2023-04-12
ACMS-000: Removed checklistapi patches.

deepakmishra2:
**Motivation**

Fixes ACMS-000

**Proposed changes**
checklistapi patches used in acquia_cms_common are now merged in the module and hence removing this fixes ACMS installation issue.

### DIFF
--- a/modules/acquia_cms_common/composer.json
+++ b/modules/acquia_cms_common/composer.json
@@ -102,9 +102,7 @@
                 "3162210 - Preview link accidentally closes the media library": "https://www.drupal.org/files/issues/2020-10-06/3162210-17.patch"
             },
             "drupal/checklistapi": {
-                "2311855: Count the default setting on the total percentage counter": "https://www.drupal.org/files/issues/2021-10-26/checklistapi-count-default-values-forprogress-2311855-2.patch",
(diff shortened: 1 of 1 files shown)


modules/acquia_cms_common/composer.json
```diff
@@ -100,11 +100,6 @@
             },
             "drupal/focal_point": {
                 "3162210 - Preview link accidentally closes the media library": "https://www.drupal.org/files/issues/2020-10-06/3162210-17.patch"
-            },
-            "drupal/checklistapi": {
```
rajeshreeputra, line 104: first patch is not available in 2.1.x, so we can not remove it.
cc @chandan-singh7929

modules/acquia_cms_common/composer.json
```diff
@@ -100,11 +100,6 @@
...
-                "2311855: Count the default setting on the total percentage counter": "https://www.drupal.org/files/issues/2021-10-26/checklistapi-count-default-values-forprogress-2311855-2.patch",
```
rajeshreeputra, line 105: first patch is not available in 2.1.x, so we can not remove it.
cc @chandan-singh7929

deepakmishra2: yup , only second patch is merged, hence added the first patch back.

rajeshreeputra:
This can be closed as patch is not required, as per latest 2.1.x release of checklistapi.

PR #1575, opened 2023-06-21
Merge branch `feature/compatible-php8.2` into `develop`

vishalkhode1:


.github/workflows/acquia_cms_ci.workflow.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           node-version: 12.13.1
       - uses: shivammathur/setup-php@v2
         with:
-          php-version: 8.1
+          php-version: 8.2
```
apathak18, line 43: would be better if CI workflow run on both PHP 8.1 and PHP 8.2

vishalkhode1: @apathak18 Yes, I'll create a separate ticket to run CI on PHP 7.4/8.0/8.1 as well.

composer.json
```diff
@@ -128,13 +128,49 @@
...
+            },
+            "drupal/schemata": {
+                "3358292 - PHP 8.2 compatibility": "https://git.drupalcode.org/project/schemata/-/merge_requests/8.patch"
+            },
+            "drupal/subrequests": {
+                "3361224 - Fix Subrequests unit tests Deprecations": "https://git.drupalcode.org/project/subrequests/-/merge_requests/16.patch"
+            },
+            "drupal/token": {
+                "3358295 - PHP 8.2 compatibility": "https://git.drupalcode.org/project/token/-/merge_requests/38.patch"
+            },
```
chandan-singh7929, line 173: I would suggest to use patch before we do actual release if these MR doesn't get merge till the time.

.github/workflows/acquia_cms_ci.workflow.yml
rajeshreeputra: +1

composer.json
rajeshreeputra: agree with @chandan-singh7929, if we are not providing patch then add there patches in release note so customer will be aware that, they need to apply these patches.

rajeshreeputra (approved):
Changes looks good to me

chandan-singh7929:
Also, one test is failing
`There was 1 failure:

1) Drupal\Tests\acquia_cms_headless\Functional\DashboardNextjsSitesTest::testSectionAdmin
Failed asserting that actual size 0 matches expected size 1.`

PR #1579, opened 2023-06-22
ACMS-1861: Fix Drupal Check notices for Drupal 10.1

rajeshreeputra:
**Motivation**
Fixes #ACMS-1861

**Proposed changes**
Fix Drupal Check notices for Drupal 10.1

**Alternatives considered**
NA

**Testing steps**
Check CI(drupal-check), it should pass.

**Merge requirements**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -170,6 +170,9 @@
             },
             "mglaman/drupal-check": {
                 "Add custom rules to bypass check for buildForm": "https://github.com/mglaman/drupal-check/pull/286.patch"
+            },
+            "stecman/symfony-console-completion": {
+                "Return type fix for stecman/symfony-console-completion": "https://github.com/stecman/symfony-console-completion/pull/94.patch"
             }
         }
     },
--- a/composer.json
+++ b/composer.json
@@ -159,9 +159,6 @@
             "drupal/scheduler": {
                 "3357836 - PHP 8.2 compatibility": "https://git.drupalcode.org/project/scheduler/-/merge_requests/85.patch"
             },
-            "drupal/schema_metatag": {
-                "PHP fatal error": "https://git.drupalcode.org/project/schema_metatag/-/merge_requests/33.patch"
-            },
             "drupal/schemata": {
                 "3358292 - PHP 8.2 compatibility": "https://git.drupalcode.org/project/schemata/-/merge_requests/8.patch"
             },
--- a/modules/acquia_cms_common/composer.json
+++ b/modules/acquia_cms_common/composer.json
@@ -13,7 +13,7 @@
         "drupal/autologout": "^1.4",
         "drupal/config_ignore": "^3.0@beta",
         "drupal/config_rewrite": "^1.5",
-        "drupal/core": "^10.0.9",
+        "drupal/core": "^10.1",
         "drupal/diff": "^1.1",
         "drupal/entity_clone": "^2.0@beta",
         "drupal/field_group": "^3.4",
@@ -76,7 +76,6 @@
         "enable-patching": true,
         "patches": {
             "drupal/core": {
-                "3059955 - It is possible to overflow the number of items allowed in Media Library": "https://www.drupal.org/files/issues/2019-12-28/3082690-80.patch",
                 "3301692: Passing null to parameter to mb_strtolower() is deprecated with PHP 8.1": "https://git.drupalcode.org/project/drupal/-/merge_requests/2598.patch"
             }
         }
--- a/modules/acquia_cms_common/src/Commands/AcmsCommands.php
+++ b/modules/acquia_cms_common/src/Commands/AcmsCommands.php
@@ -270,7 +270,7 @@ class AcmsCommands extends DrushCommands implements SiteAliasManagerAwareInterfa
   public function starterKit($filter = '', $options = [
     'project' => self::REQ,
     'format' => 'table',
-  ]) {
+  ]): bool|string|array {
     if ($starter_kit = $this->acmsUtilityService->getStarterKit()) {
       $data['starter-kit'] = $starter_kit;
       $result = new PropertyList($data);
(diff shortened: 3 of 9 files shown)


modules/acquia_cms_common/src/Services/ConfigImporterService.php
```diff
@@ -209,7 +210,7 @@ public function doImport(StorageComparer $storage_comparer, DrushLoggerManager $
         $message = 'The import failed due to the following reasons:' . "\n";
         $message .= implode("\n", $config_importer->getErrors());
 
-        watchdog_exception('config_import', $e);
+        Error::logException('config_import', $e);
```
deepakmishra2, line 213: use DI here @rajeshreeputra

deepakmishra2: as per discussion we cannot use DI here because of static method implementation in Error class similar to Crypt class

tests/src/ExistingSite/InstallStateTest.php
```diff
@@ -97,8 +97,8 @@ public function testConfig(): void {
     // Check page caching set for one year and css/js are aggregated.
     $performance_config = $this->config('system.performance');
     $this->assertSame(31536000, $performance_config->get('cache.page.max_age'));
-    $this->assertSame(TRUE, $performance_config->get('css.preprocess'));
-    $this->assertSame(TRUE, $performance_config->get('js.preprocess'));
+    $this->assertTrue(TRUE, $performance_config->get('css.preprocess'));
+    $this->assertTrue(TRUE, $performance_config->get('js.preprocess'));
```
chandan-singh7929, line 101: `assertTrue` does expect two condition, please use `$this->assertTrue($performance_config->get('css.preprocess'));` instead.

chandan-singh7929 (changes requested):
Overall changes looks good to me, there are some minor comment please check.

tests/src/ExistingSite/InstallStateTest.php
rajeshreeputra: done

apathak18 (commented):
@rajeshreeputra Can we target only on the drupal-check fixes? PHPCS fixes could be done later.

cc: @chandan-singh7929 @vishalkhode1

modules/acquia_cms_common/src/Commands/AcmsCommands.php
```diff
@@ -286,6 +286,7 @@ function ($key, $cellData, FormatterOptions $options, $rowData) {
       return $result;
     }
     $this->output()->writeln('Starter-kit is not installed yet.');
+    return FALSE;
```
apathak18, line 289: Revert this change. Not required

rajeshreeputra: It is required as this method returns data, other it will throw "return statement is missing" error in drupal-check.

composer.json
```diff
@@ -173,6 +170,9 @@
             },
             "mglaman/drupal-check": {
                 "Add custom rules to bypass check for buildForm": "https://github.com/mglaman/drupal-check/pull/286.patch"
+            },
+            "stecman/symfony-console-completion": {
+                "Return type fix for stecman/symfony-console-completion": "https://github.com/stecman/symfony-console-completion/pull/94.patch"
```
apathak18, line 175: Revert this change, Not required ~~ Merged into starterkit

modules/acquia_cms_common/composer.json
```diff
@@ -13,7 +13,7 @@
         "drupal/autologout": "^1.4",
         "drupal/config_ignore": "^3.0@beta",
         "drupal/config_rewrite": "^1.5",
-        "drupal/core": "^10.0.9",
+        "drupal/core": "^10.1",
```
apathak18, line 16: @rajeshreeputra Not sure about this change I guess "^10" would work.

modules/acquia_cms_common/src/Commands/AcmsCommands.php
apathak18: Already pushed into develop branch if you revert this change and check then it won't throw any error, also remove BOOL from return type of starterkit method

modules/acquia_cms_common/src/Services/ConfigImporterService.php
apathak18, line 213: @rajeshreeputra prior to D10.1 this would throw an undefined method error ~~ can we add version_compare logic?
cc: @vishalkhode1

rajeshreeputra: added version compare

composer.json
rajeshreeputra: it starterkit new version get released we need this.

modules/acquia_cms_common/composer.json
rajeshreeputra: As discussed in DSM, minimum requirement is need to add as ^10.1

modules/acquia_cms_common/src/Services/ConfigImporterService.php
```diff
@@ -15,6 +15,7 @@
 use Drupal\Core\Extension\ThemeHandlerInterface;
 use Drupal\Core\Lock\LockBackendInterface;
 use Drupal\Core\StringTranslation\TranslationInterface;
+use Drupal\Core\Utility\Error;
```
chandan-singh7929, line 18: Does this class available in below D10.1 version?

apathak18 (approved):

modules/acquia_cms_common/src/Commands/AcmsCommands.php
rajeshreeputra: removed.

modules/acquia_cms_common/src/Services/ConfigImporterService.php
rajeshreeputra: yes it is available in 10.0

chandan-singh7929 (approved):

vishalkhode1 (approved):

PR #1594, opened 2023-07-07
ACMS-1880: Fixing redirection login issue with moderation redirection login.

apathak18:
**Motivation**

Fixes ACMS-1880

**Proposed changes**
Adding a conditional logic if moderation dashboard exists and redirect on logic checkbox is checked.


rajeshreeputra (approved):
changes looks good to me.

deepakmishra2 (approved):

modules/acquia_cms_common/src/RedirectHandler.php
```diff
@@ -119,12 +131,14 @@ protected function handleRedirect(FormStateInterface $form_state): void {
       // is fixed.
       $this->request->query->remove('destination');
 
-      if ($this->isContributor($user)) {
-        // @todo Don't redirect if Moderation Dashboard is not enabled.
+      if ($this->isContributor($user) &&
+      $this->moduleHandler->moduleExists('moderation_dashboard') &&
+      $this->configFactory->getEditable('moderation_dashboard.settings')->get('redirect_on_login')) {
```
chandan-singh7929, line 136: I think we should use `$this->configFactory->get('moderation_dashboard.settings')->get('redirect_on_login')` instead to respect the overrides.

chandan-singh7929 (approved):
Overall changes looks good to me, though there's minor suggestion.

vishalkhode1 (approved):

modules/acquia_cms_common/src/RedirectHandler.php
apathak18: Done

chandan-singh7929 (approved):

PR #1628, opened 2023-09-04
ACMS-1920: Hide administration menu links that do not have visible child links

chandan-singh7929:
**Motivation**
For users with limited permissions (i.e. content editors), many menu items irrelevant to the role still appear in the toolbar. This can be overwhelming so empty top level menu items should have restricted access and not be shown if the user lacks the permission for it.

This is a known problem where an unnecessarily complicated interface needs to be explained/simplified to newer users.

Fixes #[ACMS-1920](https://backlog.acquia.com/browse/ACMS-1920)

**Proposed changes**
The current approach is to add a route parameter to several routes. Then add an access check for routes that have the new parameter. The access check loads the corresponding menu link and its children. If the user does not have access to any of the children, then deny access to the page. If any of the children have the new route parameter, then recursively check them.

**Alternatives considered**
N/A

**Testing steps**

- Login as the site's superuser, and create a new role - name it anything you wish
- On the 'permissions' page for the new role, grant access to 'Use the administration pages and help', 'View the administration theme' and 'Use the toolbar'
- Create a new user for the site, and give this user the newly created role
- log out as the superuser, and login as the new user, with the new role
- Notice that under 'Manage', you still have "Structure" - clicking it will take you to that page that states that you have no administrative items.

**Merge requirements**

**Screenshots**
Before:
Screenshot 2023-09-04 at 13-47-07 test_page Acquia CMS
After:
Screenshot 2023-09-04 at 13-51-14 test_page Acquia CMS

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1049,16 +1049,16 @@
         },
         {
             "name": "composer/semver",
-            "version": "3.3.2",
+            "version": "3.4.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/composer/semver.git",
-                "reference": "3953f23262f2bff1919fc82183ad9acb13ff62c9"
+                "reference": "35e8d0af4486141bc745f23a29cc2091eb624a32"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/composer/semver/zipball/3953f23262f2bff1919fc82183ad9acb13ff62c9",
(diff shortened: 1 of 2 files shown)


rajeshreeputra (approved):
Changes looks good to me.

apathak18 (approved):

vishalkhode1 (approved):

PR #1649, opened 2023-10-11
ACMS-1989: Update admin theme in Acquia CMS to Gin.

rajeshreeputra:
**Motivation**
Fixes #[ACMS-1989](https://acquia.atlassian.net/browse/ACMS-1989)

**Proposed changes**
This PR is created to check if there any test failure with Gin theme.

**Alternatives considered**
This can be closed without any further action once the issue identification is completed, or it can be used, and further changes can be made here directly.

**Testing steps**
NA

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_common/acquia_cms_common.install
+++ b/modules/acquia_cms_common/acquia_cms_common.install
@@ -27,16 +27,15 @@ function acquia_cms_common_install($is_syncing) {
       'view media',
     ]);
 
-    // Enable acquia_claro theme.
-    $themesList = \Drupal::service('extension.list.theme')->getList();
-    $themesInstall = ["olivero"];
-    if (isset($themesList['acquia_claro'])) {
-      $themesInstall[] = "acquia_claro";
-    }
-    \Drupal::service('theme_installer')->install($themesInstall);
-    \Drupal::configFactory()->getEditable('system.theme')->set('default', 'olivero')->save();
-    if (in_array("acquia_claro", $themesInstall)) {
-      \Drupal::configFactory()->getEditable('system.theme')->set('admin', "acquia_claro")->save();
+    // Enable gin theme.
+    $themes_list = \Drupal::service('extension.list.theme')->getList();
+    $theme_installer = \Drupal::service('theme_installer');
+    $theme_installer->install(["olivero"]);
+    $system_theme_config = \Drupal::configFactory()->getEditable('system.theme');
+    $system_theme_config->set('default', 'olivero')->save();
+    if (isset($themes_list['gin'])) {
+      $theme_installer->install(["gin"]);
+      $system_theme_config->set('admin', "gin")->save();
     }
 
     // Re-write the content and media view on module install,
--- a/composer.json
+++ b/composer.json
@@ -17,8 +17,7 @@
         }
     ],
     "require": {
-        "acquia/acquia-cms-starterkit": "^1.0",
-        "drupal/acquia_claro": "^1.3",
+        "acquia/acquia-cms-starterkit": "dev-feature/admin-gin-theme",
         "drupal/acquia_cms_audio": "dev-develop",
         "drupal/acquia_cms_component": "dev-develop",
         "drupal/acquia_cms_dam": "dev-develop",
--- a/composer.json
+++ b/composer.json
@@ -130,6 +130,9 @@
             "drupal/core": "-p2"
         },
         "patches": {
+            "acquia/acquia-cms-starterkit": {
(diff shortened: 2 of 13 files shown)


modules/acquia_cms_common/acquia_cms_common.install
```diff
@@ -27,16 +27,16 @@ function acquia_cms_common_install($is_syncing) {
       'view media',
     ]);
 
-    // Enable acquia_claro theme.
+    // Enable gin theme.
     $themesList = \Drupal::service('extension.list.theme')->getList();
     $themesInstall = ["olivero"];
-    if (isset($themesList['acquia_claro'])) {
-      $themesInstall[] = "acquia_claro";
+    if (isset($themesList['gin'])) {
```
apathak18, line 33: Can we do something like this?
$themes_install =  $themes_list['gin'] ? ['olivero', 'gin'] : ['olivero'];

modules/acquia_cms_common/acquia_cms_common.install
```diff
@@ -27,16 +27,16 @@ function acquia_cms_common_install($is_syncing) {
...
+      $themesInstall[] = "gin";
     }
     \Drupal::service('theme_installer')->install($themesInstall);
     \Drupal::configFactory()->getEditable('system.theme')->set('default', 'olivero')->save();
-    if (in_array("acquia_claro", $themesInstall)) {
-      \Drupal::configFactory()->getEditable('system.theme')->set('admin', "acquia_claro")->save();
+    if (in_array("gin", $themesInstall)) {
+      \Drupal::configFactory()->getEditable('system.theme')->set('admin', "gin")->save();
```
apathak18, line 39: Can we do something like this?
$system_config_factory = \Drupal::configFactory()->getEditable('system.theme');
$system_config_factory->set('default', 'olivero')->save();
if (in_array("gin", $themes_install)) {
$system_config_factory->set('admin', 'gin')->save();
}

apathak18 (approved):
Added a minor suggestion rest everything looks good

modules/acquia_cms_article/tests/src/Functional/BlogTest.php
```diff
@@ -22,7 +22,7 @@ class BlogTest extends BrowserTestBase {
   /**
    * {@inheritdoc}
    */
-  protected $defaultTheme = 'stark';
+  protected $defaultTheme = 'gin';
```
chandan-singh7929, line 25: I don't think we should test this on gin

modules/acquia_cms_common/acquia_cms_common.install
chandan-singh7929: @apathak18 - Any reason for this suggestion?

chandan-singh7929 (changes requested):
@rajeshreeputra - I have requested some changes, please take a look.

modules/acquia_cms_common/acquia_cms_common.install
apathak18: Doing the same thing with 3-4lines and variable name should be same

rajeshreeputra: @apathak18 requesting re-review.

apathak18 (approved):

deepakmishra2 (approved):

chandan-singh7929 (commented):
@rajeshreeputra - I can see some changes which are not related to work of this ticket, like refactoring, I would suggest to this in separate ticket and let only put the change which are related to this ticket.

Also don't include patch for the work that is not merged, rather provide steps to pull that in during verification.

rajeshreeputra:
> I can see some changes which are not related to work of this ticket, like refactoring, I would suggest to do this in a separate ticket and lets add only changes which are related to this ticket.
Removed the changes not related to work of this ticket.

>Also don't include patch for the work that is not merged, rather provide steps to pull that in during verification.
The patch has been added to the root composer.json file. In our CI environment, the site is installed using the starterkit, and it's important to note that there have been changes related to the Gin theme, although these changes have not been released yet. This means that our CI process will encounter failures because the starterkit version 1.1.5 now requires Acquia Claro to be installed in order to set up the site and run tests. However, in the pull request, Acquia Claro has been removed, causing the CI failures. Hence it is required to be added in root composer.json file.

chandan-singh7929:
> The patch has been added to the root composer.json file. In our CI environment, the site is installed using the starterkit, and it's important to note that there have been changes related to the Gin theme, although these changes have not been released yet. This means that our CI process will encounter failures because the starterkit version 1.1.5 now requires Acquia Claro to be installed in order to set up the site and run tests. However, in the pull request, Acquia Claro has been removed, causing the CI failures. Hence it is required to be added in root composer.json file.

ok, then lets create a ticket to remove this patch when we do the release for this changes.

PR #1650, opened 2023-10-12
ACMS-1975: Sending acms logs to sumologic.

apathak18:
**Motivation**

Fixes ACMS-1975

**Proposed changes**
Adding logger to send data to sumologic


deepakmishra2 (approved):

rajeshreeputra (approved):

modules/acquia_cms_common/src/EventSubscriber/KernelTerminate/AcquiaCmsTelemetry.php
```diff
@@ -193,12 +205,15 @@ public function sendTelemetry(string $event_type, array $event_properties = []):
       $this->sendEvent($event);
       $this->state->set('acquia_cms_common.telemetry.data', json_encode($event_properties));
       $this->state->set('acquia_cms_common.telemetry.timestamp', $this->time->getCurrentTime());
+      $this->logger->get($event_type)->info(print_r($event, TRUE));
```
chandan-singh7929, line 208: @apathak18 - I think we should also update the docblock with the change we are making here.
```
/**
   * Creates and sends an event to Amplitude.
   *
   * @param string $event_type
   *   The event type. This accepts any string that is not reserved. Reserved
   *   event types include: "[Amplitude] Start Session", "[Amplitude] End
   *   Session", "[Amplitude] Revenue", "[Amplitude] Revenue (Verified)",
   *   "[Amplitude] Revenue (Unverified)", and "[Amplitude] Merged User".
   * @param array $event_properties
   *   (optional) Event properties.
   *
   * @return bool
   *   TRUE if event was successfully sent, otherwise FALSE.
   *
   * @throws \Exception
   *   Thrown if state key acquia_telemetry.loud is TRUE and request fails.
   *
   * @see https://amplitude.zendesk.com/hc/en-us/articles/204771828#keys-for-the-event-argument
   */
```
Also, are we sure we want data to be send in array format and not in json?

apathak18: As discussed with @vishalkhode1 we are sending data in print_r format only.

modules/acquia_cms_common/src/EventSubscriber/KernelTerminate/AcquiaCmsTelemetry.php
```diff
@@ -193,12 +205,16 @@ public function sendTelemetry(string $event_type, array $event_properties = []):
       $this->sendEvent($event);
       $this->state->set('acquia_cms_common.telemetry.data', json_encode($event_properties));
       $this->state->set('acquia_cms_common.telemetry.timestamp', $this->time->getCurrentTime());
+      // Logging the database and collecting data into sumologic.
+      $this->logger->get($event_type)->info(json_encode($event_properties, JSON_UNESCAPED_SLASHES));
```
rajeshreeputra, line 209:
```suggestion
      $this->logger->get($event_type)->info('@message', ['@message' => json_encode($event['event_properties'], JSON_UNESCAPED_SLASHES)]);
```

chandan-singh7929 (approved):

modules/acquia_cms_common/src/EventSubscriber/KernelTerminate/AcquiaCmsTelemetry.php
apathak18: done

rajeshreeputra (approved):

PR #1657, opened 2023-10-24
ACMS-2006: Remove failing Geocoder patch as released in 3.35 and 4.10 version.

vipin-mittal-acquia:
Remove Geocoder patch from Acquia CMS Place

**Motivation**
This will solve failing ACMS Installation
 
**Proposed changes**
Remove geocoder patch

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -2800,12 +2800,12 @@
             "dist": {
                 "type": "path",
                 "url": "./modules/acquia_cms_place",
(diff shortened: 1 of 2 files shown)


chandan-singh7929:
We should also update module minimum version from where this path has been included.

rajeshreeputra:
> We should also update module minimum version from where this path has been included.

Updated the module minimum version to latest.

rajeshreeputra (approved):

chandan-singh7929 (approved):

PR #1673, opened 2023-11-07
ACMS-1997: Adding request id  with ACMS data for sumologic.

apathak18:
**Motivation**

Fixes ACMS-1997

**Proposed changes**
Adding request_id along with ACMS Data in sumologic


vishalkhode1:
This is not required. Hence, closing this.

PR #1708, opened 2023-12-18
ACMS-000: Update workflow to test on Drupal Core 10.2.x.

vishalkhode1:


composer.json
```diff
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "acquia/acquia-cms-starterkit": "^1.0",
+        "acquia/acquia-cms-starterkit": "1.1.5",
```
rajeshreeputra, line 20: Any reason to pin starterkit to 1.1.5?

rajeshreeputra (changes requested):

PR #1712, opened 2023-12-21
ACMS-3362: Updating packages alter.

apathak18:
**Motivation**

Fixes ACMS-3362

**Proposed changes**
Validating support of contenthub 3.3.x and CI should pass.

### DIFF
--- a/acquia-pipelines.yml
+++ b/acquia-pipelines.yml
@@ -5,7 +5,7 @@ services:
   - composer:
       version: 2
   - php:
-      version: 8.1
+      version: 8.2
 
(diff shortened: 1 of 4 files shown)


acquia-pipelines.yml
```diff
@@ -5,7 +5,7 @@ services:
   - composer:
       version: 2
   - php:
-      version: 8.1
+      version: 8.2
```
deepakmishra2, line 8: i think we should also test this on 8.3

deepakmishra2 (approved):

PR #1717, opened 2024-01-03
ACMS-3474: Move specific modules from info.yml to .install file.

apathak18:
**Motivation**

Fixes ACMS-3474

**Proposed changes**
Moved out smart_trim and moderation_sidebar from .info.yml file

### DIFF
--- a/modules/acquia_cms_common/acquia_cms_common.install
+++ b/modules/acquia_cms_common/acquia_cms_common.install
@@ -83,7 +83,6 @@ function acquia_cms_common_install($is_syncing) {
     // Contributed modules shipped by Acquia CMS.
     $contributed_modules = [
       'entity_clone',
-      'smart_trim',
       'moderation_sidebar',
       'syslog',
       'purge_processor_cron',
--- a/modules/acquia_cms_common/acquia_cms_common.install
+++ b/modules/acquia_cms_common/acquia_cms_common.install
@@ -67,12 +67,6 @@ function acquia_cms_common_install($is_syncing) {
       $module_installer->uninstall(['page_cache']);
     }
 
-    // Install entity clone module during common installation.
-    if (!$module_handler->moduleExists('entity_clone')) {
-      $module_installer->install(['entity_clone']);
-      $config_storage->write('entity_clone.settings', $source_optional_dir->read('entity_clone.settings'));
-    }
-
     /*
      * We've to explicitly set purge plugin as acquia_purge, If Acquia Purge
      * module is enabled, else this would give below error:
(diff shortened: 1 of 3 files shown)


modules/acquia_cms_common/acquia_cms_common.install
```diff
@@ -86,12 +80,18 @@ function acquia_cms_common_install($is_syncing) {
...
+      'smart_trim',
+      'moderation_sidebar',
       'syslog',
       'purge_processor_cron',
       'purge_ui',
-    ]);
+    ];
+    // Install the syslog module and purge sub-modules.
+    // And other contributed modules provided by Acquia CMS.
+    $module_installer->install($contributed_modules);
```
rajeshreeputra, line 94: @apathak18, I believe it would be more appropriate to relocate this code from the .install file to a helper or service file and then invoke the method from there. What are your thoughts on this, @vishalkhode1?

modules/acquia_cms_common/acquia_cms_common.module
```diff
@@ -381,6 +383,28 @@ function acquia_cms_common_modules_installed($modules, $is_syncing) {
...
+      // Iterate optional configs to be included for entity view display.
+      foreach ($modules_suffix as $suffix) {
+        $acms_module = "acquia_cms_$suffix";
+        // Check if module exists.
+        if ($module_handler->moduleExists($acms_module)) {
+          _acquia_cms_common_rewrite_configuration("core.entity_view_display.node.$suffix.teaser", $acms_module);
+          _acquia_cms_common_rewrite_configuration("core.entity_view_display.node.$suffix.search_results", $acms_module);
+        }
+      }
+    }
```
rajeshreeputra, line 407: @apathak18, I believe it would be more appropriate to relocate this code from the .install file to a helper or service file and then invoke the method from there. What are your thoughts on this, @vishalkhode1?

apathak18: But it's in .module file

rajeshreeputra: updated the file extension in comment.

modules/acquia_cms_common/acquia_cms_common.install
apathak18: Are we repeating this logic somewhere?

chandan-singh7929 (commented):
@apathak18 - Overall changes looks good to me , but I have once concern, since we are removing hard dependency of smart_trim. what will happen if customer uninstall this module from the codebase.
Considering the usage in CT.
image

PR #1740, opened 2024-02-16
ACMS-3561: ACMS modules to use generic template.

apathak18:
**Motivation**

Fixes ACMS-3561

**Proposed changes**
Use generic template to run gitlab CI


deepakmishra2:
I think we no more require this, hence closing this.

PR #1750, opened 2024-02-23
ACMS-3577: DAM integration with starterkit (low code) doesn't work as expected.

deepakmishra2:
… expected.

**Motivation**

Fixes #ACMS-3577

**Proposed changes**
updated attributes for media DAM

**Testing steps**
follow from ticket

### DIFF
--- a/modules/acquia_cms_dam/config/pack_acquia_cms_dam/cohesion_templates.cohesion_content_templates.media_acquia_dam_image_asset_emb.yml
+++ b/modules/acquia_cms_dam/config/pack_acquia_cms_dam/cohesion_templates.cohesion_content_templates.media_acquia_dam_image_asset_emb.yml
@@ -169,8 +169,7 @@ json_values: |
                       ]
                   },
                   "attributes": {
-                      "alt": "[media:acquia_dam_alt_text:value]",
-                      "title": "[media:name:value]"
+                      "alt": "[media:acquia_dam_alt_text:value]"
                   }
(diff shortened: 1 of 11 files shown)


apathak18 (approved):

chandan-singh7929 (approved):
Changes looks good to me, I think we should keep same behaviour across all starterkit, see my comment on the ticket for more details.

Also, we need to think how we are going to provide this changes for our existing customers.

vishalkhode1 (approved):

modules/acquia_cms_dam/config/pack_acquia_cms_dam/cohesion_templates.cohesion_content_templates.media_acquia_dam_image_asset_emb.yml
```diff
@@ -169,8 +169,7 @@ json_values: |
                       ]
                   },
                   "attributes": {
-                      "alt": "[media:image:alt]",
-                      "title": "[media:image:title]"
+                      "alt": "[media:acquia_dam_alt_text:value]"
```
chandan-singh7929, line 172: @deepakmishra2 - This will only be generated if there is a mapping for alt text field, otherwise it will be blank because default DAM mapping doesn't have alt text field mapping out of the box.
Hence, I would suggest to use field condition then display the fallback as file name as alt value.
cc @vishalkhode1

PR #1765, opened 2024-03-08
ACMS-3560: Support moderation dashboard major release version 3.x.

deepakmishra2:
**Motivation**

Fixes #ACMS-3560

Requirements
3.0.x
D10 ONLY, dropped support for charts v3 that hasn't had a release since 2022.

**Proposed changes**

update moderation dashboard to version 3.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,8 @@
         "drupal/sitestudio_config_management": "dev-develop",
         "drupal/sitestudio_gin": "^1.0",
         "drush/drush": "^10 || ^11 || ^12",
-        "mnsami/composer-custom-directory-installer": "^2.0"
+        "mnsami/composer-custom-directory-installer": "^2.0",
+        "nnnick/chartjs": "^4.4"
     },
     "require-dev": {
         "acquia/coding-standards": "^1.0",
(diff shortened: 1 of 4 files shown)


chandan-singh7929:
Tests are failing @deepakmishra2

composer.json
```diff
@@ -37,7 +37,8 @@
         "drupal/sitestudio_config_management": "dev-develop",
         "drupal/sitestudio_gin": "^1.0",
         "drush/drush": "^10 || ^11 || ^12",
-        "mnsami/composer-custom-directory-installer": "^2.0"
+        "mnsami/composer-custom-directory-installer": "^2.0",
+        "nnnick/chartjs": "^4.4"
```
apathak18, line 41: Do we need this package in the require section of root composer.json?

deepakmishra2 (commented):
Looks good to me

vishalkhode1 (approved):

composer.json
deepakmishra2: @chandan-singh7929 - i think we need to have 3.9 and 4.4 both, for the lower version to support

rajeshreeputra: Agree with @apathak18 this is not required as acquia_cms_common has this dependency.
cc @vishalkhode1 @chandan-singh7929

PR #1773, opened 2024-03-12
ACMS-3650: Support 2.x version for node revision delete.

apathak18:
**Motivation**

Fixes ACMS-3650

**Proposed changes**
Update acquia_cms_site_studio composer.json with  2.x version of node revision delete.

### DIFF
--- a/modules/acquia_cms_site_studio/acquia_cms_site_studio.module
+++ b/modules/acquia_cms_site_studio/acquia_cms_site_studio.module
@@ -212,33 +212,30 @@ function acquia_cms_site_studio_modules_installed($modules, $is_syncing) {
       $config = \Drupal::configFactory()->getEditable('node_revision_delete.settings');
       if ($config) {
         $config->set('defaults.amount.status', TRUE)
-          ->set('defaults.amount.settings.amount', '50')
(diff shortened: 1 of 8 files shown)


modules/acquia_cms_site_studio/acquia_cms_site_studio.module
```diff
@@ -206,6 +207,67 @@ function acquia_cms_site_studio_modules_installed($modules, $is_syncing) {
...
+            ],
+          ],
+          'drafts' => [
+            'status' => TRUE,
+            'settings' => [
+              'age' => 12,
+            ],
+          ],
+        ],
+      ];
```
deepakmishra2, line 235:
```suggestion
      ];
```
@apathak18 can this code be used in a method rather than calling this in multiple places.

apathak18: No as these are the settings related to only node revision delete and facade we created for processing the config settings and third party settings in a generic way.
For basically using the facade to pass the settings as a parameter so make the structure more robust.

chandan-singh7929 (approved):

vishalkhode1 (approved):

PR #1802, opened 2024-04-15
Merge branch `stage` into `main`

vishalkhode1:


modules/acquia_cms_common/composer.json
```diff
@@ -55,7 +55,8 @@
         "drupal/acquia_cms_starter": "<1.3.3",
         "drupal/acquia_cms_toolbar": "<1.4",
         "drupal/acquia_cms_tour": "<1.3.2 || ~2.0.0",
-        "drupal/acquia_cms_video": "<1.4"
+        "drupal/acquia_cms_video": "<1.4",
+        "drupal/acquia_connector": "<4.0.6"
```
vishalkhode1, line 59: @rajeshreeputra Looks like this is is no longer required as Acquia CMS no longer uses/override any connector service.

composer.json
```diff
@@ -37,7 +37,8 @@
         "drupal/sitestudio_config_management": "dev-develop",
         "drupal/sitestudio_gin": "^1.0",
         "drush/drush": "^10 || ^11 || ^12",
-        "mnsami/composer-custom-directory-installer": "^2.0"
+        "mnsami/composer-custom-directory-installer": "^2.0",
+        "nnnick/chartjs": "^4.4"
```
rajeshreeputra, line 41: Can we remove this from root composer.json as it's already available in Acquia CMS COmmon module as dependency?

modules/acquia_cms_common/composer.json
rajeshreeputra: Yes Acquia CMS Common no longer uses any connector service but it is required on acquia server to connect.

tests/ci/script.sh
```diff
@@ -30,22 +30,19 @@ if [ "${ACMS_JOB}" == "backstop_tests" ] || [ "${ACMS_JOB}" == "upgrade_modules"
...
-  # npm run backstop-starter && error=false || error=true
-  # if [ "${error}" = "true" ]; then
-  #    tar -cvzf test_reports.tar.gz ./tests/backstop/bitmaps_test/;
-  #    aws s3 cp --recursive "./tests/backstop/bitmaps_test/" "${AWS_S3_BUCKET_PATH}/backstop/logs/${GITHUB_RUN_ID}"
-  # fi
+  npm run backstop-starter && error=false || error=true
+  if [ "${error}" = "true" ]; then
+    if [ "${COPY_BACKSTOP_IMAGES}" == "true" ]; then
+      tar -cvzf test_reports.tar.gz ./tests/backstop/bitmaps_test/;
+      aws s3 cp --recursive "./tests/backstop/bitmaps_test/" "${AWS_S3_BUCKET_PATH}/backstop/logs/${GITHUB_RUN_ID}"
```
rajeshreeputra, line 37: Any reason we are sending backstop test failure images to AWS all time.

vishalkhode1: We are not sending all the time. Default value for `COPY_BACKSTOP_IMAGES ` is false: https://github.com/acquia/acquia_cms/pull/1802/files#diff-4eb666b493503cabe7a4f72501ebd718bc71bc8560928ffc0b9b9d031f15c8e1R230
Developer can just set this value as true if they need to copy to AWS and see.

rajeshreeputra: awesome 🙌

composer.json
vishalkhode1: Let's create a separate PR for this.

PR #1849, opened 2024-06-26
ACMS-3931: Adding D11 version.

apathak18:
**Motivation**

Fixes #ACMS-3931

**Proposed changes**
Added Drupal 11 version.

### DIFF
--- a/modules/acquia_cms_tour/src/Plugin/AcquiaCmsStarterKit/StarterKitSelectionForm.php
+++ b/modules/acquia_cms_tour/src/Plugin/AcquiaCmsStarterKit/StarterKitSelectionForm.php
@@ -53,12 +53,10 @@ class StarterKitSelectionForm extends AcquiaCmsStarterKitBase {
    * {@inheritdoc}
    */
   public function buildForm(array $form, FormStateInterface $form_state) {
-
-    $service = $this->starterKitService;
     $missingModules = [
-      'acquia_cms_enterprise_low_code' => $service->getMissingModules('acquia_cms_enterprise_low_code'),
-      'acquia_cms_community' => $service->getMissingModules('acquia_cms_community'),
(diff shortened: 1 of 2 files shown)


modules/acquia_cms_tour/src/Plugin/AcquiaCmsStarterKit/StarterKitSelectionForm.php
```diff
@@ -46,7 +54,7 @@ public function getFormId() {
    */
   public function buildForm(array $form, FormStateInterface $form_state) {
 
-    $service = \Drupal::service('acquia_cms_tour.starter_kit');
+    $service = $this->starterKitService;
```
rajeshreeputra, line 57: Since we already possess a service variable, `$this->starterKitService`, I would like to understand the rationale for initializing a separate instance. Could you please clarify the necessity for this additional service variable?

```suggestion
```
@vishalkhode1 what's your opinion?

rajeshreeputra (changes requested):

modules/acquia_cms_tour/src/Plugin/AcquiaCmsStarterKit/StarterKitSelectionForm.php
apathak18: Just a avoiding the direct drupal call as we already doing these approach in the similar form i.e StarterKitConfigForm

rajeshreeputra: I mean we don't need `$service` variable directly you can use `$this->starterKitService`.

apathak18: Done

rajeshreeputra (approved):

vishalkhode1 (approved):

PR #1863, opened 2024-07-23
ACMS-4044: Initial changes for D11 Acquia cms headless.

apathak18:
**Motivation**

Fixes ACMS-4044

**Proposed changes**
Supporting D11.

### DIFF
--- a/modules/acquia_cms_headless/acquia_cms_headless.info.yml
+++ b/modules/acquia_cms_headless/acquia_cms_headless.info.yml
@@ -2,7 +2,7 @@ name: "Acquia CMS Headless"
 package: 'Acquia CMS'
 type: module
 description: 'Provides functionality for Progressively Decoupled and Purely Headless sites using Node and Next JS.'
-core_version_requirement: ^9.4 || ^10
+core_version_requirement: ^9.5 || ^10 || ^11
 dependencies:
   - acquia_cms_tour:acquia_cms_tour
   - consumers:consumers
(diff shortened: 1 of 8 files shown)


deepakmishra2 (approved):

modules/acquia_cms_article/tests/src/ExistingSite/BreadcrumbTest.php
```diff
@@ -88,12 +88,13 @@ public function testBreadcrumb(string $node_type, string $sub_type, array $expec
    * @return array[]
    *   Sets of arguments to pass to the test method.
    */
-  public function providerNoSubType() : array {
+  public static function providerNoSubType() : array {
+    $object = new self('test');
```
vishalkhode1, line 92: @apathak18 I think there's no need to create an object for this. As we are calling method `providerBreadcrumb` and this method is also static, we can simply call it like: `self::providerBreadcrumb()`. Please revisit and update at all places.

vishalkhode1 (changes requested):

modules/acquia_cms_article/tests/src/ExistingSite/BreadcrumbTest.php
apathak18: This was already merged into the base branch ~~ Now my commits are clean.

vishalkhode1 (approved):

PR #1877, opened 2024-09-13
ACMS-4210: Removing google_analytics.

apathak18:
**Motivation**

Fixes ACMS-4210

**Proposed changes**
Removing google_analytics and replace with google_tag.


rajeshreeputra (approved):

PR #1986, opened 2026-06-12
ACMS-6125:Clear Facet Filters block not rendering in Query String URL…

rohansinhadev:
… processor mode

**Motivation**

The "Clear Facet Filters" block only renders when the facet source uses the Pretty Paths URL processor. When switched to the default Query String processor, the block returns empty and the "Clear filter(s)" link never appears.

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

Add the "Clear Facet Filters" block to a Site Studio View and keep Query String processor in facets, the Clear Facet Filters will work after these changes.

**Merge requirements**

### DIFF
--- a/modules/acquia_cms_search/src/Plugin/Block/ClearFacetFilters.php
+++ b/modules/acquia_cms_search/src/Plugin/Block/ClearFacetFilters.php
@@ -75,11 +75,19 @@ class ClearFacetFilters extends BlockBase implements BlockPluginInterface, Conta
    * {@inheritdoc}
    */
   public function build() {
-    // Check if any facets_query is present in the url. If yes then display the
-    // reset link.
+    // Pretty paths mode: facets are encoded in the {facets_query} route param.
     if ($this->routeMatch->getParameter('facets_query')) {
-      $url = $this->routeMatch->getRouteName();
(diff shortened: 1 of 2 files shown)


modules/acquia_cms_search/src/Plugin/Block/ClearFacetFilters.php
```diff
@@ -75,11 +75,19 @@ public static function create(ContainerInterface $container, array $configuratio
    * {@inheritdoc}
    */
   public function build() {
-    // Check if any facets_query is present in the url. If yes then display the
-    // reset link.
+    // Pretty paths mode: facets are encoded in the {facets_query} route param.
```
rajeshreeputra, line 78: Please confirm if this fixes issue in both facets v2 and v3? 
If it is not affecting both, we might need to consider dropping support for v2 or looking for a workaround.

rohansinhadev: sure, checking on this.

rohansinhadev: hi @rajeshreeputra , its working with both facets versions.

rajeshreeputra (approved):
Looks good.

